PR #4, opened 2015-01-06
Stream converter

hasimpson-usgs:


jmfee-usgs:
Since "e" is already obtained during the conversion, always include it in the returned stream.  Make include_d the option (since it requires additional conversion):
https://github.com/usgs/geomag-algorithms/pull/4/files#diff-0f820c7ca43702648b9d8fd945ca75c1R146

jmfee-usgs:
In StreamConverter_test.py, make stats local to the __get_trace method.  Make decbas an optional parameter.  npts shouldn't need to be set, and comments should be a list/tuple instead of a string (but not sure it's needed).

hasimpson-usgs:
We wrote ChannelConverter to always consider "obs" as h,e.  The only way to
save processing time,  is to
1) change what ChannelConverter considers "obs",
2) explicitly call the lower level routines.
3) add a channelConverter routine that passes back h,d and optionally e.

I think of the 3,  I prefer 3.

On Mon, Jan 26, 2015 at 1:10 PM, Jeremy Fee <notifications@github.com>
wrote:

> Since "e" is already obtained during the conversion, always include it in
> the returned stream. Make include_d the option (since it requires
> additional conversion):
>
> https://github.com/usgs/geomag-algorithms/pull/4/files#diff-0f820c7ca43702648b9d8fd945ca75c1R146
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/usgs/geomag-algorithms/pull/4#issuecomment-71528849>.
>



-- 
Hal Simpson
IT Specialist
USGS Geomagnetism Program
Office:303-273-8459
Email:hasimpson@usgs.gov

jmfee-usgs:
This line
https://github.com/usgs/geomag-algorithms/pull/4/files#diff-68dee2c8cd6672559652262259de0a27R108

Needs to call ```parse_file``` instead of ```_parse_url```.

hasimpson-usgs:
Changed,  and pushed.

PR #5, opened 2015-01-26
Xyz

hasimpson-usgs:


jmfee-usgs:
- move ```script_dir``` inside if where it is used
- make convert_stream take arguments (stream, informat, outformat)
- use elif in convert_stream, so it will act more like a switch statement
- move main docstring to top of file (after #!)
- reading from a file doesn't require a url, or an absolute path:
```
iagaFile = open(args.infile, 'r').read()
```
- reading from stdin is faster without the loop:
```
iagaFile = sys.stdin.read()
```
- make obs with d a different in/out format ("obsd"?)
- don't provide defaults for informat/outformat, they should be required
- rename factory method ```parse_file``` to ```parse_string```, update doc string.
- add factory method ```write_string``` so main script only needs to use factory, update put_timeseries to use new method

jmfee-usgs:
xyz.py
- line 29, this file is not named main.py
- move `from os import path` inside the if (only place it's used).
- move `import sys` outside the if, since it is also used in main()
- Calls to StreamConverter.get_obs_from_geo should use named parameters for the boolean values, and you only need to include the parameter that overrides the default (i.e. include_d=True).
- if we support `obsd` as an output format, we should also support it as an input format (read in the file that was written).
- close fh after writing if it's a file that was opened, or change to use `with` similar to in IAGA2002Factory

IAGA2002Factory
- change write_string method to write_file since it accepts a file object as a parameter.

hasimpson-usgs:
Pushed latest up.

jmfee-usgs:
get_out_channels
- is this method still needed?

convert_stream
- previous comment about include_e was for outformat=='obs' and informat=='obs(d)'
- this is a little hard to follow, split to check only outformat in outer if then check informat in inner if:
```
if outformat == 'geo':
  if informat == 'obs' or informat == 'obsd':
    ...
elif outformat == 'mag':
  ...
```

check_stream:
- loop over the channels that are expected, and throw an exception if that channel isn't found, something like:
```
for channel in channels:
  if len(timeseries.select(channel=channel)) == 0:
    return False
return True
```

hasimpson-usgs:
Pushed latest up.  Not certain I did the right thing on check_stream,  but your code was cleaner then mine.

jmfee-usgs:
- update comment at top of file for informat/outformat obs vs obsd.
- in convert_stream, outformat == 'obs', informat == 'obs' or informat == 'obsd', `include_d` should be `include_e` (in case informat is obsd and outformat is obs).
- remove blank lines in convert_stream method, indentation separates sections
- informat, outformat should be required (add `, required=True` to `add_argument` call).
- when reading from stdin, you may want to print a message to that effect to stderr (in case someone runs interactively and wonders why it's "hanging").

PR #6, opened 2015-02-13
Docs

hasimpson-usgs:
This is being done as an experiment.  Dumped the google doc to html,  then converted it online via http://domchristie.github.io/to-markdown/ to markdown.   It seems to work.  Although the comments in the google doc,  got appended to the bottom.   
Let me know your thoughts,  it's a bit clunky,  but I think it will work until we have something more formal figured out.


jmfee-usgs:
Clean up the markdown
- rename to XYZ.md
- named anchors are not needed, github adds them
- spans are not needed
- update tables to use markdown
- use definition list for HDZ, XYZ
- reference "geo", "mag", and "obs" as used within xyz library
- update diagram to include all of geo/mag/obs vectors in different colors and line styles (to make sure they are grayscale safe)
- move images to an xyz folder, rename to figureX....
- add alt text to images
- for images containing equations, are images needed in the first place?

jmfee-usgs:
Much better than the original auto generated markdown.  Some additional suggestions:

- Add a space between the # and heading (`# Heading`)
- Remove extra blank lines between sections, there should be 1.
- Wrap lines at 80 characters, in sublime you can show the ruler under `View->Ruler->80` and wrap using `Edit->Wrap->Wrap Paragraph at Ruler`.
- Remove the HTML entities (`&rsquo;`, `&nbsp;`)


Background and Motivation
- Use code ticks around H, D, Z, X, Y, Z when they are being defined, use common punctuation and grammar.  For X,Y,Z, use similar language as for H,D,Z; Maybe "`X` is the magnitude of the geographic north pole component of the H vector".


Math and Theory
- Use code ticks around equations and label equation numbers, also add anchors and link references:
```
- <a name="eq1"></a>Equation 1: `X = H cos(D)`
```
```
Eqs [1](#eq1), [2](#eq2), [3](#eq3)
```
- "from HD to he" doesn't provide the equation for little e.
- in the "potential for mathematically undefined" comment that internally we use the atan2 variant to handle this.


Algorithm
- it may be worth mentioning that the channels are still all uppercase (or maybe we should use lower case) for observatory data.


Practical Considerations
- Some of these have been addressed by the library and we should mention those specifics:
- "##Declination Angular Units".  The library internally uses radians for all angles, and factories convert into this standard.  You can optionally reference the ChannelConverter.get_radians_from_minutes.
- "##Declination Baseline".  The library IAGA factory attempts to parse DECBAS from the iaga comments section.  You can optionally link to metadata for where users can find this.
- "##Data Flags".  The library internally represents data gaps as NaN, and factories convert to this where possible.

hasimpson-usgs:
Incorporated latest suggestions.

jmfee-usgs:
This has been merged.

PR #16, opened 2015-03-09
Added EdgeFactory, ObservatoryMetadata and their accompanying tests.  Ed...

hasimpson-usgs:
...geFactory only has get_timeseries implemented, put_timeseries will be done later.  ObserveatoryMetadata has the Observatory information hardcoded for now, future implementations will use something more dynamic.


jmfee-usgs:
EdgeFactory
- add an optional observatory metadata parameter to __init__ to allow custom configuration (default `None`), create an instance of ObservatoryMetadata if none provided
- in get_timeseries, only check for a default `channels` value once
- initialize timeseries to be an empty stream and append channels (`timeseries = obspy.core.Stream()`)
- call timeseries.merge before returning stream
- the put_timeseries NotImplementedError should say "put_timeseries"
- `get_edge_channel_codes` is not used
- `get_interval_from_edge` is not used
- is `get_channel_code_from_edge` needed? already know channel that was requested.
- is `_get_edge_code_from_channel` needed? appeasr to already be implemented in `_get_edge_channel`.
- is `_get_interval_from_code` needed? already know interval that was requested.
- create a `_set_metadata(stream, observatory, channel, type, interval)` method in EdgeFactory that for now calls ObservatoryMetadata instance.  Update stats on all traces in stream (all for the same channel anyways, but may have gaps).  Also set `channel`, `type`, and `interval`.
- update the existing stats object instead of creating a new stats object based on the existing.


EdgeFactory_test
- add methods to test component methods `_get_edge_network`, `_get_edge_station`, `_get_edge_channel`, `_get_edge_location`
- add method to test `_set_metadata`
- remove tests for removed methods from above.


ObservatoryMetadata
- default metadata should be all caps: `DEFAULT_METADATA`
- update formatting of default
-- opening curly braces on same line as equals or colon
-- closing curly braces on own line
-- continuation lines indented twice
- `channel` isn't needed
- `data_interval`, `data_interval_type`, `data_type` should vary (possibly set in EdgeFactory)
- should `is_intermagnet` be true?
- add `__init__` that accepts optional metadata reference.  If `None`, make instance use DEFAULT_METADATA.
- look up observatory based on set_metadata parameter
- set all values in stats from metadata reference
- `.keys()` is not needed:
```
for key in metadata:
  stats[key] = metadata[key]
```
or
```
for key, value in metadata.iteritems():
  stats[key] = value
```
- don't return stats, just update stats that were passed in


ObservatoryMetadata
- add test using custom metadata, verify custom declination_base returned instead of default.

jmfee-usgs:
EdgeFactory
- `get_type_from_edge` and `get_interval_from_code` aren't needed
- comment for `type` parameter to other _get_edge... methods lists title cased values, should be lower case (`definitive`, `quasi-definitive`, `variation`).
- remove commented out lines from _get_timeseries
- move `timeseries.merge()` from get_timeseries to _get_timeseries before calling _set_metadata
- remove prints from _set_metadata

ObservatoryMetadata
- constant should be upper-snake-case `DEFAULT_METADATA`
- remove unused keys, which overwrite valid data: `network`, `station`, `channel`, `data_type`, `data_interval`, `data_interval_type` (possibly set this per observatory, but based on data_interval value).
- comments shouldn't begin with `#`, the iaga writer adds that.
- `is_intermagnet = True` generates the last two lines of the comments (Final data ... on obtaining this product.)
- `condtions_of_use` should be `conditions_of_use` (note the i between d and t)
- The `This data file was constructed by the Golden GIN` comment should probably be added by the iaga writer, and not the edge metadata.
- The check for None can be simplified:
```
    self.metadata = metadata or DEFAULT_METADATA
```
- set_metadata should accept observatory, channel, type, and interval parameters
- move stat update for channel, data_interval, and data_type into ObservatoryMetadata
- `data_interval_type` also needs to be set based on whether data_interval is minute or second (currently metadata only sets `filtered 1-minute (00:15-01:45).

jmfee-usgs:
edge/__init__.py
- not based on ngdc.noaa.gov url...

EdgeFactory
- simplify how observatoryMetadata is set:
`
self.observatoryMetadata = observatoryMetadata or ObservatoryMetadata()
`
- `_get_timeseries` returns `obspy.core.Stream` instead of trace.


ObservatoryMetadata
- remove `data_interval_type` from DEFAULT_METADATA
- is there a reason to use `100 Hz` and `0.01 second`?  can we pick one?
- will different observatories have different `data_interval_type`s?
- update each metadata to be more future proof:
`
OBS: {
  data_interval_type: {
    'minute': '...',
    'second': '...'
  },
  metadata: {
    ...
  }
}
`
- use DEFAULT_DATA_INTERVAL_TYPE if observatory doesn't specify custom, make data interval types a parameter to the ObservatoryMetadata constructor (same as metadata).
- don't check if key already exists, metadata should overwrite.


IAGA2002Parser
- `is_gin` should default to False (can never be false currently)

IAGA2002Writer
- is_gin should generate "Golden GIN" instead of "GoldenGIN"

docs/metadata.md
- add `is_gin` key to metadata reference

hasimpson-usgs:
Asking Duff about 100 Hz vs. 0.01 sec. 
data_interval_type(s) can be different.  Tim has played around with a filtered 1-second avg on obs-rio.

jmfee-usgs:
this is now #18

PR #18, opened 2015-03-27
Xyz algorithm

hasimpson-usgs:
Just code for now.  Need to add comments (I was pretty bad), and unit tests.   Will move files out of conf, and into test/Example directory after they're sanitized.


jmfee-usgs:
Some initial comments

bin/geomag.py
- don't need to document arguments in the script, just document the help option and that method parses arguments to configure a Controller.
- should only be one call to StreamIAGA2002Factory for input, one for output, whether stream is stdin/stdout or a file.

geomagio/Algorithm.py
- separate input_channels and output_channels, have subclass provide them in call to init (based on config)

geomagio/Controller
- require an algorithm, geomag.py always creates one.

hasimpson-usgs:
Changes made.

jmfee-usgs:
delete geomagio/WaveserverFactory, Edge replaces this.
make test files end in _test.py


bin/__init__.py
- is this so you can import geomag for testing?


bin/geomag.py
- what is the `#noqa` comment?
- use `add_argument_group` to separate "input", "algorithm", and "output" arguments.
- use `add_mutually_exclusive_group` where it makes sense; like ("input-edge", "input-iaga-file", "input-iaga-magweb", "input-iaga-stdin", "input-iaga-url")
- make input-edge include 2 positional arguments: host and port.
- require entire iaga-url in argument (instead of 3 separate arguments); remove _get_iaga_output_url, _get_iaga_input_url methods.
- update output arguments to work simlar to input
- use named parameters when creating xyz algorithm
- pass a file or stdin object into the StreamIAGA2002Factory (similar to how you do for writing), have factory `read()` from object.


bin/main.py
- import sys should only happen in `except` block when importing geomagio
- what is the `#noqa` comment?


geomagio/XYZAlgorithm.py
- update `check_stream` to use `self.inchannels`
- update `check_stream` to raise an Exception (create an `AlgorithmException` class) when input channel not found in timeseries.
- call `check_stream` at start of process


geomagio/edge/EdgeFactory.py
- remove `edge_add_parse_arguments` method
- consolidate code that is cleaning up data, `_convert_trace_to_decimal` is called in `_get_timeseries` but `_post_process` is called in `get_timeseries`.  Do all cleanup in one place so all the steps involved are clearer.


geomagio/edge/ObservatoryMetadata.py
- do the filter comments (for example BOU.metadata.filter_comments) make sense for all intervals? or should those be grouped with data_interval_type?
```
'data_interval_type': {
  'minute': {
    'name': 'Filtered 1-minute (00:15-01:45)',
    'comments': [
      # ...
    ]
  },
  'second': {
    'name': 'Average 1-second',
    'comments': []
  }
}
```
- I think you want to prefer observatory specific data_interval information:
```
data_interval_type = self.data_interval_type
if 'data_interval_type' in self.metadata[observatory]:
  data_interval_type = self.metadata[observatory].data_interval_type
# use data_interval_type
```


geomagio/iaga2002/StreamIAGA2002Factory
- did slice return an extra data point without the `interval` argument?

hasimpson-usgs:
bin __init__.py is needed for lint and grunt watch to run on the directory.  Also if we want to write test files. 
#noqa is needed for lint to ignore the warning that geomagio isn't used. 
I'm using WaveServerFactory to help test edge_out.   I'll delete it later.

hasimpson-usgs:
Changes made, per request.

jmfee-usgs:
Looks good, mainly coding standard issues:

bin/geomag.py
- argument parser will already check the length of args.input_edge
- two levels of indentation on continuation lines, near `inputfactory = edge.EdgeFactory`
- don't read from input_iaga_file or input_iaga_stdin; pass file into StreamIAGA2002Factory and have it call read.
- check the input arguments in alphabetical order
- check the output arguments in alphabetical order
- add arguments within each section to parser in alphabetical order
- make metavar's upper case `HOST`, `PORT`

geomagio/edge/__init__.py
- trailing comma after `'ObservatoryMetadata'`

geomagio/iaga2002/StreamIAGA2002Factory.py
- call parse_string with `self._stream.read()`, similar comment in bin/geomag.py; stream should always be a stream object and not a string for input.

gruntconfig/flake8.js, gruntconfig/watch.js
- put different src, files values on separate lines:
```
src: [
    'geomagio/**/*.py',
    'bin/**/*.py'
]
```

setup.py
- missing comma before `'geomagio.edge'`

PR #26, opened 2015-05-07
Edge out

hasimpson-usgs:
Using the "write all channels/sleep/close" methodology.


hasimpson-usgs:
After fix to edge,  converted to instantiate and close RawInputClient on every channel,  instead of each complete dataset.

jmfee-usgs:
First round of comments:

EdgeFactory.py
- put_timeseries should still accept starttime and endtime arguments to be consistent with TimeseriesFactory interface.  When None, they can default to timeseries start/end.
- put_timeseries should use trace metadata as defaults for observatory, type, interval...
- self.ric is only used in _put_trace, convert to local variable
- _put_trace, prefer location over self.locationCode
- move most of _put_trace to RIC.  EdgeFactory should 1) create RIC, 2) convert trace to int, 3) call RIC.send_trace
- reference `ma` as `numpy.ma` to be more explicit about where it comes from

RawInputClient.py
- does not need to edit path, can assume geomagio is on the path already.
- add comment for MAXINPUTSIZE global
- can tag be generated by RawInputClient? instead of configured?
- remove self.(cwbhost, cwbport, cwbsocket, now, dummy, buf)
- remove separate cwbhost/cwbport/cwbsocket; this complexity should go elsewhere.
- remove now/dummy timestamps, just say UTCDateTime(datetime.utcnow()) in forceout
- create `send_trace` method that accepts trace and does most logic of EdgeFactory._put_trace, EdgeFactory shouldn't need to know how RIC works.
- send()
-- update send method parameter documentation to match coding standard
-- can the array of samples be passed in and the nsamp computed?
-- can the header bet created, and then if there are samples data appended to buf to reduce duplication?
-- don't catch socket errors, need to communicate to calling code there was a problem.
- if unable to connect in _open_socket, will it try forever? maybe try 2-3 times?
- _get_next_tag()
-- build tag directly:
```
tag = self.tag + str(self.tagseq) + '     '
```
-- it may be less confusing to pass six zeros instead of using the array.
-- is it okay to send tagseq 100? checking >99 after sending may not do what you expect.

WaveserverFactory.py
- remove this file.

jmfee-usgs:
EdgeFactory
```
- This check should happen in _get_timeseries (if channel is D, then...)
+        if 'D' in channels:
+            for trace in timeseries.select(channel='D'):
+                trace.data = ChannelConverter.get_radians_from_minutes(
+                    trace.data)
```
- remove debugging `print trace_send.data)


RawInputClient
- remove comments about cwbhost/port
- move RawInputClient.Sequence outside of class and use underscore + all caps (_SEQUENCE); same for Seedname
- consider making _get_tag() and send() work similarly; separate _get_forceout(), _get_data() methods for different packet types?
- can you always include samples in the packStr? even if it's zero length?
- should open socket also send the tag?
- should the tag use a sequence/random number to avoid collisions from multiple processes?

jmfee-usgs:
EdgeFactory
- change `self.ric` to local variable instead of class variable.
- move call to `_convert_stream_to_masked` inside `_put_channel`
- make `_convert_stream_to_masked`, `_convert_trace_to_int` return copies; leaving the parameter unchanged.
- be consistent when computing sampling_rate for `_create_missing_channel`, there are values `1.`, `1 / 60.`, `1 / 3600`, etc. always use period or never.  Also, add zero after decimal point to be more explicit.


RawInputClient
- make _send accept the buffer to be sent
- have `forceout` call `_get_forceout` and pass result to _send.
- have `send_trace` call `_get_data` and pass result to _send.


Fix code convention problems:
```
>> geomagio/edge/EdgeFactory.py ... fail
>> geomagio/edge/EdgeFactory.py:162:23: E711 comparison to None should be 'if cond is None:'
>> bin/geomag.py ... fail
>> bin/geomag.py:177:36: E231 missing whitespace after ','
```

PR #30, opened 2015-06-30
Readme

emcwhirter-usgs:


jmfee-usgs:
named anchors are not needed, github automatically adds these to headers (mouseover, then click link to left of header)

PR #40, opened 2015-07-13
Forceout is now an optional part of edge out,  with the default being…

hasimpson-usgs:
… false

### DIFF
--- a/geomagio/Controller.py
+++ b/geomagio/Controller.py
@@ -274,7 +274,8 @@ def main(args):
                 type=args.type,
                 interval=args.interval,
                 locationCode=locationcode,
-                tag=args.output_edge_tag)
+                tag=args.output_edge_tag,
+                forceout=args.forceout)
     else:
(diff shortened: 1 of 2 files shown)


jmfee-usgs:
Make the argument `--output-edge-forceout` since it is (currently) only relevant for outputting to edge.

hasimpson-usgs:
I'm blaming brain fog.   I thought I should change that,  then promptly
forgot that thought.

On Mon, Jul 13, 2015 at 1:50 PM, Jeremy Fee <notifications@github.com>
wrote:

> Make the argument --output-edge-forceout since it is (currently) only
> relevant for outputting to edge.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/usgs/geomag-algorithms/pull/40#issuecomment-121036322>
> .
>



-- 
Hal Simpson
IT Specialist
USGS Geomagnetism Program
Office:303-273-8459
Email:hasimpson@usgs.gov

PR #48, opened 2015-08-24
Imfv283

hasimpson-usgs:


hasimpson-usgs:
#49

jmfee-usgs:
Initial comments:

IMFV283Factory.py
- `IMFV283_FILE_PATTERN` doesn't seem to be used.
- `read_url` should probably be refactored into a utility module.
- `get_timeseries` seems to ignore starttime/endtime (it passes current time to `_get_url`).

IMFV283Parser.py
- `ObservatoryMetadata` should be refactored out of the `edge` module.
- `IMFV_DEAD` is already in the imfv module, rename to `DEAD_VALUE`
- don't abbreviate `HDR_SIZE`, make it `HEADER_SIZE`; same for other globals
- in `_get_data`, it may be more efficient to apply scale/offset/bias/shift using numpy array operations; only replace dead values with `NaN`.
- should the parser assign arbitrary names to the channels (like '1', '2', '3', '4' when orientation is 3, so data can be accessed after parsing?
- in `_process_ness_block`, unpack the 3 bytes into 2 local variables, swap if needed, then set within goes_block.


imfv283_codes.py
- consider looping over `OBSERVATORIES` to build a lookup table for `PLATFORMS`; from platform to observatory code.

readme_io.md
- Does `--input-imfv283-url` need a placeholder (`URL`)?

jmfee-usgs:
Url
- any methods that raise `TimeseriesFactoryException`s should probably be moved to the `TimeseriesFactory` base class (with underscore prefixes since they are "internal" methods)
- `read_url` appears to be the only generic method, so I recommend moving it into the `Util` module

IMFV283Parser
- move obspy Trace and Stream creation into the Factory; the parser should deal with the input format only, while the factory should translate into the obspy format.
- replacing dead values is much more efficient as a numpy array operation, instead of per value, see https://github.com/usgs/geomag-algorithms/blob/master/geomagio/iaga2002/IAGA2002Parser.py#L171
- inline the code for `_get_data_offset`, which is only used in one place.

imfv283_codes
- the value in `PLATFORMS` should be just the object `obs` instead of a tuple `(obs)`

hasimpson-usgs:
Done, per our conversation.

jmfee-usgs:
Remove the URL module, `read_url` should move to `Util` and the remaining methods should move to `TimeseriesFactory`.

PR #51, opened 2015-11-11
Dst

hasimpson-usgs:
Added the ability to read dst data from edge,  and write it to IAGA2002.

### DIFF
--- a/geomagio/Controller.py
+++ b/geomagio/Controller.py
@@ -367,7 +367,7 @@ def parse_args(args):
             choices=['R0', 'R1', 'RM', 'Q0', 'D0', 'C0'])
     parser.add_argument('--interval',
             default='minute',
-            choices=['minute', 'second'])
+            choices=['hourly', 'minute', 'second'])
     parser.add_argument('--update',
             action='store_true',
             default=False,
--- a/geomagio/ObservatoryMetadata.py
+++ b/geomagio/ObservatoryMetadata.py
@@ -458,11 +458,11 @@ DEFAULT_METADATA = {
         },
         'interval_specific': {
             'minute': {
-                'data_interval_type': '1-minute calculated ',
+                'data_interval_type': '1-minute calculated',
                 'filter_comments': []
             },
             'hourly': {
-                'data_interval_type': '1 hour calculated'
+                'data_interval_type': '1-hour calculated'
             }
         }
     },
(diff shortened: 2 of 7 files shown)


jmfee-usgs:
Extra space in `data_interval_type`, also be consistent in using or omitting hyphens (`1-minute` vs `1 hour`):
https://github.com/usgs/geomag-algorithms/pull/51/files#diff-62bd410ad5e6ec2a7927a9b7426b6249R461

Should `channel` be allowed to fall through if it isn't known, for future channels?
https://github.com/usgs/geomag-algorithms/pull/51/files#diff-417434e38ce5a3adec5c917100ec8046R355

Move `_create_empty_trace` to a utility class
https://github.com/usgs/geomag-algorithms/pull/51/files#diff-fc63f2d0948d8f4df114d719d495df37R39

Should we allow up to 3 characters, instead of specifically 1 or 3?
https://github.com/usgs/geomag-algorithms/pull/51/files#diff-fc63f2d0948d8f4df114d719d495df37R194

Use a different "empty" channel name that doesn't include spaces (instead of ` - `), maybe `NUL`?
https://github.com/usgs/geomag-algorithms/pull/51/files#diff-fc63f2d0948d8f4df114d719d495df37R251

Did you verify the IAGA2002Parser can read files with channel names that are longer than 1 character?
Make sure the IAGA2002Parser drops the "empty" channels.

hasimpson-usgs:
IAGA2002Parser does indeed deal fine with longer then 1 character channels.  
IAGA2002Parser does NOW drop empty channels. 
All other changes made per suggestion.

jmfee-usgs:
- Only import the necessary classes from obspy
https://github.com/usgs/geomag-algorithms/pull/51/files#diff-421e9081b22caacd028fed7c3c622becR3

instead of `import obspy`, use `from obspy.core import Stats, Trace`


- Update the docstring parameters
https://github.com/usgs/geomag-algorithms/pull/51/files#diff-421e9081b22caacd028fed7c3c622becR61


- This looks like it would consider any channel with the same value as empty:
https://github.com/usgs/geomag-algorithms/pull/51/files#diff-0b32b2423a25c72dae4d2222f42b63f5R170

It may work better by moving it after replacing dead values with `numpy.nan` and checking the count of nan.  However, it's possible for a legitimate channel to have a gap, so it's probably best to only remove "empty" channels that use the empty channel name `NUL`.

PR #71, opened 2016-01-27
Usgs/geomag algorithms#64

### LINKED ISSUE #64: add sq/dist algorithm readme
link from relevant documentation places

emcwhirter-usgs:
Fixes #64 add sq/dist algorithm readme
Fixes usgs/geomag-algorithms#64

Here is a first attempt at a simplified readme.


jmfee-usgs:
Define the components more like how XYZ does:
https://github.com/usgs/geomag-algorithms/blob/master/docs/algorithms/XYZ.md

Be more explicit that this implementation is using holt winters...

PR #78, opened 2016-01-29
Monitor

hasimpson-usgs:


jmfee-usgs:
- Currently has an error:
```
$ bin/monitor.py 
Traceback (most recent call last):
  File "bin/monitor.py", line 321, in <module>
    main(args)
  File "bin/monitor.py", line 252, in main
    if args.report_type == hourly:
NameError: global name 'hourly' is not defined
```

- Print usage if missing required arguments (or make arguments required and argparse will handle)

- A text format would be more useful on the command line.
- Have a non-zero exit code if there are "problems", too many gaps, etc.  This would support integration into icinga.

hasimpson-usgs:
Merge gitlab #6 when this is merged.

jmfee-usgs:
Looks okay for now, recommend separating the logic for getting current observatory status, determining whether to warn, and html output formatting.

PR #80, opened 2016-02-01
Describe SqDist parameters more efficiently.

emcwhirter-usgs:


geomagio/algorithm/SqDistAlgorithm.py
```diff
@@ -584,15 +584,15 @@ def add_arguments(cls, parser):
         """
         parser.add_argument('--sqdist-alpha',
                 default=1.0 / 1440.0 / 30,
-                help='SqDist alpha parameter',
+                help='Smoothing parameter for secular variation',
                 type=float)
         parser.add_argument('--sqdist-beta',
                 default=1.0 / 1440.0 / 30,
-                help='SqDist beta parameter',
+                help='Smoothing parameter for slope, default 0',
```
jmfee-usgs: The line above this suggests the default is actually `1.0 / 1440.0 / 30` instead of 0.

jmfee-usgs:
Based on discussion with Josh:
- change default value for beta to 0
- remove default from argument help

PR #83, opened 2016-02-02
Usgs/geomag algorithms#76

### LINKED ISSUE #76: make sqdist algorithm use state
If the first requested time doesn't match the next_starttime SqDistAlgorithm property, the state is considered invalid and discarded (get_input_interval and process_one methods).

option, when using state:
- do not process any data before next_starttime.  In get_input_interval, if endtime is before next_starttime, don't clear the state but also in process_one return an empty stream.
- in get_input_interval, if the requested starttime is after next_starttime (instead of equal) return an interval that starts at next_starttime.

emcwhirter-usgs:
Fixes #76 make sqdist algorithm use state

### DIFF
--- a/geomagio/algorithm/SqDistAlgorithm.py
+++ b/geomagio/algorithm/SqDistAlgorithm.py
@@ -49,6 +49,32 @@ class SqDistAlgorithm(Algorithm):
         self.next_starttime = None
         self.load_state()
 
+    def can_produce_data(self, starttime, endtime, stream):
+        """Determine wheter data can be produced.
+
+        Parameters
+        ----------
+        starttime: UTCDateTime
+            start time of requested output.
+        endtime : UTCDateTime
+            end time of requested output.
+        stream: obspy.core.Stream
+            The input stream we want to make certain has data for the
+            algorithm.
+        """
+        # check state
+        if self.last_observatory is not None \
+                and self.last_channel is not None \
+                and self.next_starttime is not None:
+            # Check for (and skip) older gaps when a state is present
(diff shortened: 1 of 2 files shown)


jmfee-usgs:
self.next_starttime is not defined for all Algorithms.

A SqDistAlgorithm.can_produce_data method should be implemented, and can call the base Algorithm.can_produce_data for the normal interval checking

emcwhirter-usgs:
In this case do we still want to clear the state if it is not valid, and then check where the state file is in time?

jmfee-usgs:
- in clear_state, don't check if self.statefile is None; always clear internal state.  save_save will return if statefile is none.

- in can_produce_data, the call to Algorithm.can_produce_data when there is state is redundant (remove, it will fall through to call below).  Change to `return Algorithm.can_produce_data`...

- in get_input_interval, len(channels) == 1 seems like it should be above (with/before channels[0] == last_chanel).  Checking end < start seems like it should compare end < next_starttime; and if that works always return (next_starttime, end), right?

- in process_one, call clear state if state is not in sync; update handling in can_produce_data and get_input_interval will ensure data _should_ line up, and manual calls should just not use state.

jmfee-usgs:
The tests are failing, my guess is it's the last check since the stream is empty.

emcwhirter-usgs:
It was actually just a whitespace error that didn't show up until I restarted grunt. They're all passing now.

jmfee-usgs:
I think you also need to check the stream contents, i.e. do the stream traces all start at next_starttime  (even if a gap)

PR #85, opened 2016-02-03
Usgs/geomag algorithms#77

### LINKED ISSUE #77: allow algorithms that process more data than originally requested to return that data as well
The controller currently trims the processed data to the requested start/end time.  This is only an issue in the previous case if you pad the starttime to be next_starttime; the controller may trim the result and not save the SQ/SV/Dist channels between next_starttime and requested starttime.

option:
- add a command line argument (trim by default, allow override like --no-trim) and update controller run method to conditionally trim the processed data before output.

emcwhirter-usgs:
Fixes #77

### DIFF
--- a/geomagio/Controller.py
+++ b/geomagio/Controller.py
@@ -159,7 +159,9 @@ class Controller(object):
                     timeseries=timeseries,
                     renames=options.rename_input_channel)
         processed = algorithm.process(timeseries)
-        processed.trim(starttime=options.starttime, endtime=options.endtime)
+        # trim if --no-trim is not set
+        if not options.trim:
+            processed.trim(starttime=options.starttime, endtime=options.endtime)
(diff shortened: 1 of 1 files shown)


jmfee-usgs:
There is no `options.trim`, this should be `options.no_trim`

geomagio/Controller.py
```diff
@@ -159,7 +159,10 @@ def run(self, options):
                     timeseries=timeseries,
                     renames=options.rename_input_channel)
         processed = algorithm.process(timeseries)
-        processed.trim(starttime=options.starttime, endtime=options.endtime)
+        # trim if --no-trim is not set
+        if not options.no_trim:
+            processed.no_trim(starttime=options.starttime,
```
jmfee-usgs: The stream method is still `trim`, only the option is `no_trim`

PR #89, opened 2016-02-11
Add plot output factory

jmfee-usgs:
For displaying data as the output.


emcwhirter-usgs:
Remove unused `os` and `TimeseriesFactoryException` imports from `PlotTimeseriesFactory.py`

PR #93, opened 2016-02-12
91

### LINKED ISSUE #91: Implement recursion limit for update flag.

hasimpson-usgs:
fixes #91

### DIFF
--- a/bin/monitor.py
+++ b/bin/monitor.py
@@ -269,7 +269,7 @@ def print_observatories(args):
             print summary_table
             print gap_details
 
-        return warning_issued
+    return warning_issued
 
 
 def main(args):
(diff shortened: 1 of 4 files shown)


hasimpson-usgs:
This adds a flag to limit the number of recursion steps in update.

jmfee-usgs:
- remove debug print, or add debug option that disables it by default.
- change update_count to be an optional parameter to the run_as_update method, rather than a controller attribute, defaulting to zero.
- fix linting errors.

hasimpson-usgs:
Changed and fixed per notes above.

jmfee-usgs:
- self._update_count is not needed
- increment update_count when recursing (self.run_as_update(options, update_count + 1)), just return if at limit

hasimpson-usgs:
Made suggested changes.

Issue #47, opened 2015-08-05
Implement Adjusted data


emcwhirter-usgs:
This requires further discussion before implementation.

From Jira:

> This data is calculated from the output of the geomag-baseline-calculator and applied to the raw data stream. One of the standard algorithms for this is to do a simple average of all of the baseline points for the last 4 weeks, but this will need to be configurable (a linear fit and a least squared algorithm can also be used for different situations).

Related e-mail from Bill Worthington:

> Adjusted data are produced by using the last 4 to 8 weeks of absolute measurements to project a preliminary baseline forward in time to produce the adjusted data in real-time.  At some observatories, where Absolutes are observed frequently, a least-squared fit is used to fit the baseline values and project them forward.  For observatories where the Absolutes are less frequent, a simple average of the last 4 baseline values are used to project the baselines forward in time.

There is also a relevant paper titled:

> On the feasibility of promptly producing quasi-definitive magnetic observatory data

PR #103, opened 2016-04-06
Phase 1 of addressing Issue #47.  Needs more tests, to be added later

aclaycomb-usgs:
more docstring/documentation/commenting needed.


jmfee-usgs:
Some initial comments:
- delete unused imports instead of commenting out
- in `__init__`, pass `('H', 'E', 'Z', 'F')` for inchannels and `('X', 'Y', 'Z', 'F')` for outchannels to Algorithm.__init__
- in `load_state`, the second call `self.matrix = np.eye(4)` is redundant with the call before loading the statefile
- in `load_state`, do the matrix elements need to be `np.array`s? or would it be more direct to use `np.float64`?
- at the end of `process`, update the created traces (or possibly override Algorithm.create_trace) to set each channels `stats.data_type` to `adjusted` and `stats.location` to `A0`; make the target data_type and location configurable parameters in the contructor and arguments

And to complete the integration with the Controller and Documentation:
- "register" the algorithm in algorithm/__init__.py, by adding to both the `algorithms` and `__all__` variables
- add a documentation page in `docs/algorithms`; both background and usage.  Add links from documentation/api.md and documentation/usage.md

jmfee-usgs:
Adjusted.md
- you mention `z` is downward in a couple places, but not as part of the definition of `Z`.
- do `H_absolute` and `H_ordinate` need sqrt?
- `offset error` has open paren `(` but no closing paren?
- line wrapping is inconsistent

Adjusted_usage.md
- Add example statefile content, include statefile argument in initial usage (since without state algorithms is identity transform)

AdjustedPhaseOneFunction2.ipynb
- maybe add text to explain what is happening in different steps through the example

api.md
- list in alphabetical order (Adjusted first)

usage.md
- adjusted header without content?

AdjustedAlgorithm.py
- if you create an instance with a specific matrix and pier_correction, but no statefile, they are overwritten by load_state.
- In load_state, I recommend letting any exceptions be thrown; something went wrong so it shouldn't continue as if it didn't
- In create_trace, `Trace` should be lowercase; make data_type and location code configurable via constructor

algorithm/__init__.py
- use alphabetical order

AdjustedAlgorithm_test
- test_process should run without requiring a network connection; save a data file in IAGA/other format in the etc folder and use that.
- more tests

PR #106, opened 2016-04-27
added temp, vbf factories and custom chan.loc in edge factory

dcstewart-usgs:
Factories to support temperature, vbf, binlog and 1-sec pcdcp files


jmfee-usgs:
Included in https://github.com/usgs/geomag-algorithms/pull/109

PR #109, opened 2016-05-11
Dcstewart usgs new factories plus edits

jmfee-usgs:
Includes Duff's factory changes.
Adds support for url intervals other than 1 day.
Simplifies input/output factory arguments
Makes IAGA2002 writer more flexible
Adds IMFV122 input factory

Replaces https://github.com/usgs/geomag-algorithms/pull/106


jmfee-usgs:
The new Util.read_url method uses pycurl, so that will need to be installed.

For sftp support on OSX I needed to reinstall curl using brew before installing pycurl:
```
brew install curl --with-libssh2
brew link --force curl
```

Then install pycurl using pip
```
pip install pycurl
```

PR #120, opened 2016-07-08
Goes password

hasimpson-usgs:
Added a password option for goes.

### DIFF
--- a/geomagio/Controller.py
+++ b/geomagio/Controller.py
@@ -299,6 +299,7 @@ def get_input_factory(args):
         input_factory = imfv283.GOESIMFV283Factory(
                 directory=args.input_goes_directory,
                 getdcpmessages=args.input_goes_getdcpmessages,
+                password=args.input_goes_password,
                 server=args.input_goes_server,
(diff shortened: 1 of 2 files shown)


jmfee-usgs:
Remove the new prints:
https://github.com/usgs/geomag-algorithms/pull/120/files#diff-a8061c4c1cfad85fdbba675ed206709dR128

This seems unrelated:
https://github.com/usgs/geomag-algorithms/pull/120/files#diff-eb4ccf738235c6cfed6c4639b24600daR584

jmfee-usgs:
Since this is an array of arguments:
https://github.com/usgs/geomag-algorithms/pull/120/files#diff-a8061c4c1cfad85fdbba675ed206709dR130

Shouldn't the argument flag and value be separate array entries?

Like this:
```
[self.getdcpmessages,
                      '-h ', server,
                      '-u ', self.user,
```

Instead of:
```
[self.getdcpmessages,
                      '-h ' + server,
                      '-u ' + self.user,
```

PR #122, opened 2016-07-25
Do not try to set metadata when observatory is not known

jmfee-usgs:


aclaycomb-usgs:
travis says geomagio/ObservatoryMetaData.py, line 999 that in the change reads:

if not observatory in self.metadata:

should read:

if observatory not in self.metadata:

PR #123, opened 2016-07-27
IPython notebook updates

erigler-usgs:
- updated Ipython notebook to used SqDistAlgorithm class and methods
- added unit tests
- added functional tests/demonstrations
- added lots of explanatory Markdown


erigler-usgs:
I made substantial changes to the SqDist.ipynb Notebook (including change its name) in the "release candidate" branch I created as per Michelle Guy's guidelines for peer review. I now need a technical/scientific review of the SqDist Algorithm, as well as an administrative review for coding style, etc. I name @aclaycomb-usgs for the technical review, and @jmfee-usgs for the admistrative review. Please use GitHub to comment on this pull request.

jmfee-usgs:
Update the link to the ipython notebook to reference the new location:
https://github.com/usgs/geomag-algorithms/blob/5d2292f210ec57d66a20c08f6ca030e7fa65bab8/docs/algorithms/SqDist.md#L56

I know most of this has been developed to deal specifically with minute data.  Not sure if you want to touch on how this should be applied depending on the data sampling period and intended usage, i.e. `m=1440` for minutes of a day, `beta=0`, etc.?

erigler-usgs:
I updated SqDist.md to point to the renamed Notebook (see 0ad6965).

In response to comment/question by @jmfee-usgs, we might want to consider changing the default alpha and gamma parameters to zero, which will make the algorithm agnostic to the sampling period. If we do this, we need to make sure we are setting these explicitly in the production scripts, and not relying on default behavior.

jmfee-usgs:
I think either approach is okay, and they currently default to the most common use so I'd vote to leave the defaults unchanged.  The effect of parameters should still be discussed in documentation even if only to say "they default to minute data with a seasonal interval of 1 day".

erigler-usgs:
I think we do want to change defaults to zero, and not just because this makes it agnostic to the sampling period. 

It implies that that the baseline level (SV) and/or seasonal correction (SQ) are supposed to remain fixed, a very reasonable assumption. This is evident if/when the user specifies only alpha, or only gamma...more than likely they intended for the parameter that was *not* specified to equal zero.

The sensible default for beta is not so obvious, but we already decided that it should be zero, so I'm inclined to leave it as-is. Same for phi, except that its default should remain one.

Changes are here (58c85f2); corresponding changes must be made to production scripts on GitLab before this pull request can be merged.

jmfee-usgs:
Please add a new pull request to change the defaults, or include those changes in this pull request.  Also describe what values the production scripts should be configured to use.

erigler-usgs:
Maybe I'm misunderstanding something @jmfee-usgs , but are the commits added to this pull request (see 58c85f2) not sufficient to change the command line defaults?

As for the production scripts, I'll talk to @aclaycomb-usgs about changing those on GitLab, but the updated SqDist_usage.md file gets at the gist of it.

jmfee-usgs:
I missed that that commit was already added to this PR...  Do you think it's worth describing the relation between 1/30 and 1/30/1440? or allowing the 1, 30, and 1440 to be parameterized directly?

erigler-usgs:
I definitely think it is worth describing how the "forgetting factors" work in more detail. But I also think a lot of the Geomag Algorithms documentation needs to be refactored, updated, and just generally polished up, so I was considering a separate branch and pull request to deal with all of this. For now, I just want to get SqDist approved and into production.

As for your second part of your comment, I think you are proposing we parameterize what I would call a "memory", not a "forgetting factor". It's all semantics of course, but most of the literature I've seen, and everything I've cited, refers to the "forgetting factor". Also, it's a little more straight-forward to specify zero (0) forgetting than it is to specify infinite (Inf?) memory from the command line.

jmfee-usgs:
How about as a compromise describing the recommended settings for minutes data more explicitly around here, and parameterization or examples for other data intervals can be added later:
https://github.com/usgs/geomag-algorithms/pull/123/files#diff-1a2f79f1458f08570c6775c7b4f602f2R54

erigler-usgs:
I polished up the SqDist_usage.md file (see 80e29db). Hopefully things are a bit clearer now.

erigler-usgs:
I added more explanatory detail to SqDist.md (see 5795f99)

jmfee-usgs:
SqDistValidate.ipynb

- I like the suggestion to move the interface information to the usage page.
- The class should be `geomagio.algorithm.SqDistAlgorithm` (note the lower case `algorithm`)
- I would separate the `additive` and `estimate_parameters` methods into a separate 'Class Methods' section
- If Test 1 - 5 are the unit tests, you could refer to the file that contains them, rather than repeat them in the notebook.

SqDist.md

I still think a summary list introducing the parameters `m`, `alpha`, `beta`, and `gamma` would be useful (similar to how `Dist`, `SQ`, and `SV` are introduced).  The example values could also be in a separate summary beneath the description of these parameters.  Something like:

```
Exponential Smoothing

parameter list

discussion of parameters

recommended values for 1 minute data
```

erigler-usgs:
@jmfee-usgs,

I think I've addressed all, or most of your concerns. I moved the API description from the Notebook to the SqDist_usage.md file. I suspect we might separate this into something like SqDist_cli.md and SqDist_api.md in the future, but for now this should work as a basic reference. It's all in my latest commit (7cbcfff).

As for the unit tests you mentioned, those in the Notebook are changed considerably from the ones you (or Eddie?) originally set up, and that are the /test directory. They actually use the geomagio.algorithm.SqDistAlgorithm class and member functions rather than call additive() directly. They may or may not be used as unit tests, but they should remain in the Notebook as known working demonstrations for users.

jmfee-usgs:
I think this is close, a few more suggestions for the background document

docs/algorithms/SqDist.md
- Rather than saying "we won't discuss here" you could link to the "Why Exponential Smoothing" section or move the Why section up on the page
- Include `m` in the parameter list
- When describing `gamma=1/43200*1440`, it would be more explicit to use parenthesis:
```
gamma=(1/43200)*1440
```
- Is there a reason to choose 30 days, and cycles of 1 day?  Please briefly describe why these are chosen at the beginning of the minute data example, possibly introducing it as our best practice, "For 1-minute timeseries data, USGS assumes a 30 day secular variation period and 1 day cycle because ..., which results in parameter values of ..."

erigler-usgs:
- I actually tried to provide a (provably overly) concise explanation of the reasons why traditional Fourier methods don't perform well in real-time processing, so there is no longer a "we won't discuss here" statement.

- I added `m` to the parameter list

- I don't agree that the parentheses make anything more clear in the sample `gamma` definition, and my coding style preference to to avoid unnecessary parentheses.

- I inserted a parenthetical statement motivating the 30 day time scale. However, I don't want to go so far as to actually recommend a setting, since a major point of the algorithm is to give the user some flexibility. Indeed, we eventually want these settings to be optimized on a per-observatory basis.

aclaycomb-usgs:
Technical reveiw:

- Does the split signal add up to the original signal? If not, by how much are the two different?
 
- There are some 'high frequency' information and unit-step-function-like signals still in SV/SQ.  Is there a plan to smooth that, or transfer it to Dist?  If so, does that change the total energy (power) content of SV + SQ + Dist?  by how much?

- If there is a storm, what will be the distribution of the storm signal across the three parts?  

- What can we expect as the user changes the forgetting factors, damping etc?

erigler-usgs:
@aclaycomb-usgs 

I'm just going to respond to your questions here (pull request #123), and NOT submit another pull request to modify the Jupyter Notebook. My answers are relatively short, and I just discovered that I can insert a figure to help make some of my points.

1) The split signal does indeed add up to the original signal.

2) As of this pull request, the step-function like signals I believe you are referring to (i.e., the ones noted in proto-production runs) still exist. These arise when there is a relatively large disturbance, and assuming that they are indeed a transient phenomenon, they get smoothed exponentially. For the SV signal, they disappear fairly quickly, since SV is updated with every observation. For SQ, they tend to linger because each element of SQ only gets updated once-per-day (or more precisely, once per sample-cycle, or every 1440 1-minute observations).

Anyway, to answer your first question, yes, there is a plan to smooth SQ by applying a local boxcar filter, which will help mitigate these transients in SQ. The following snippet demonstrates how such a modification might be applied to SqDistAlgorithm.py (here s represents SQ, and et represents Dist):

```python
# original s update
# s[i + m] = s[i] + gamma * (1 - alpha) * et

# s updated with 1-hour centered boxcar smoothing
s[i + m] = s[i] + gamma * (1 - alpha) * et / 61.
s[i+m-30:i+m] = s[i+m-30:i+m] + gamma * (1-alpha) * et / 61.
s[i+1:i+30+1] = s[i+1:i+30+1] + gamma * (1-alpha) * et / 61.
```

To answer your second question, there is no change in the power content of SV+SQ+Dist for the very reason given in item 1 above. Of course the power in SV, SQ, and Dist individually will change, but this is impossible to predict unless we know the exact amplitude and duration of these step functions.

3) Ideally, if there is a storm, there will be no impact on SV or SQ, since Dist will exceed the Z-score threshold (multiplied by a time-dependent standard error), thus preventing updates to SV and SQ. The following figure demonstrates this for the March 17, 2015 storm:

smooth_bou_days_000to180

The light gray envelope represents the Z-score threshold. If the blue line falls outside of this, SV and SQ are not updated. Of course the Z-score threshold needs to be chosen carefully, and will probably require a different setting for each observatory. The objective is to update SV and SQ only during relatively quiet times, but of course if you set the Z-score threshold too low, little or no smoothing will ever occur.

4) Smaller forgetting factors (i.e., alpha, beta, gamma) mean that the average age of data contributing to the current estimate of SV or SQ increases, so more data points are being smoothed. This is made evident in Equation (2) in the Jupyter Notebook, and I attempted to clarify things in an updated version of [docs/algorithms/SqDist.md](https://github.com/usgs/geomag-algorithms/blob/master/docs/algorithms/SqDist.md). But perhaps a more intuitive explanation can be found at this link [http://people.duke.edu/~rnau/411avg.htm](http://people.duke.edu/~rnau/411avg.htm) (scroll down to section that begins "The average age of the data in the simple-exponential-smoothing forecast is 1/α...")

jmfee-usgs:
@erigler-usgs 

re 1) split signal adding to original signal, is this true even in the case of spikes that are beyond Xsigma?


re 4), I think Abe is asking for similar clarification of the effects of alpha, beta, gamma.  It may be worth referencing that link in the readme.

erigler-usgs:
@jmfee-usgs 

re 1) yes
re 4) not sure I understand what you are saying or suggesting here

jmfee-usgs:
@aclaycomb-usgs can you clarify what you were asking about for 4?  My interpretation is he is looking for additional information on how alpha, beta, and gamma, affect SQ/Dist.  It still isn't clear to me why/how I would change those values for a desired effect.

aclaycomb-usgs:
For 4, Josh's answer prompted me to go back through the SqDist.md, and I think that the section titled Exponential Smoothing is sufficient for someone versed in signal processing or smoothing/averaging scientific time series data.  I do see some value in linking to a tutorial on exponential smoothing, for potential users who have not seen this before.   

I was thinking something like:  Increasing alpha effects the SV signal by decreasing the amount of time it takes for SV to move from a stable value through a rapid change in the original signal to a new stable value of SV, provided the average of the original signal has that step change, with relatively little change before or after.  Similarly for gamma, the higher gamma is, the shorter the time scale is for changes to the SQ signal.   Since these are time-discrete signals, sample rate and time period drive the value for alpha, beta, gamma, combined with the time scale desired for discounting old information in the running average.   alpha = 1/(samples_per_period*number of periods) ; gamma = (samples_per_cycle)/(samples_per_period*number of periods).  For SV, the weight of the difference in the new value and previous value is alpha.  Alpha, beta, gamma only make sense in this context to be between 0 and 1.  

I also realize, especially after writing the above, that it's hard in a paragraph to convey that.  We can't teach semester-long classes in the documentation, but pointing people to the resources to get up to speed is a minimum.

This is just my understanding above, based on reading about this algorithm and from discussions in the office.  I may be wrong.  Really, it comes down to the audience for this algorithm and its users.  Primarily, this will be an operational, real-time data producing tool, and the references provided should be sufficient for the geomag science community to understand what is being done to the data.  I'm somewhat playing devil's advocate by asking (4), but with the goal that we understand what we're using and developing around.

aclaycomb-usgs:
Apparently I can't put `*` in by itself.  

`alpha = 1/(samples_per_period*number of periods)`

`gamma = (samples_per_cycle)/(samples_per_period*number_of_periods)`

erigler-usgs:
Submitted a new pull request (#129) to continue reconciling these reviews. Hopefully it does not go as long as this one.

PR #125, opened 2016-08-09
Update docs

jmfee-usgs:
Add a dockerfile, and update README.md to be a little friendlier.
The docker image has yet to be published, so that portion will not work yet.


jmfee-usgs:
The docker image has now been pushed.

PR #126, opened 2016-08-18
Update update

jmfee-usgs:
- Eliminate a redundant call to get_input_timeseries when in update mode.
- Update empty values used by IAGA2002 and IMFV122 factories


aclaycomb-usgs:
travis says undefined symbol: clock_gettime

PR #133, opened 2016-10-20
SqDist local smoothing

erigler-usgs:
I implemented local smoothing for SQ to address previously noted issues with both real and artificial high-frequency noise that could not be addressed efficiently via the standard exponential smoothing parameter gamma. This required some relatively small changes to SqDistAlgorithm.additive(), including a new configuration parameter. It also required related changes to the command-line options.

I don't believe this change is substantial enough to warrant a full and formal peer review. However, I tested things on 6-months of 2003 data (includes the so-called Halloween Storm) from 3 different observatories, and everything worked as expected. I'm attaching representative snapshots of the BOU results that demonstrate just what this new parameter does when a "cutoff period" of 3 hours (180 1-minute samples) is specified.

bou_solarquiet009

bou_solarquiet009

I discussed these positive results with both Abe Claycomb and Tim White.  We all agreed that the choice to use a Gaussian filter over a boxcar filter was appropriate. Tim, not unexpectedly, encouraged using a more sophisticated filter. I might do this in the future, but for now the Gaussian filter meets our somewhat loosely defined requirements in terms of spectral response.

### DIFF
--- a/geomagio/algorithm/SqDistAlgorithm.py
+++ b/geomagio/algorithm/SqDistAlgorithm.py
@@ -471,7 +471,7 @@ class SqDistAlgorithm(Algorithm):
                 s[i + 1:i + nts / 2 + 1] = (s[i + 1:i + nts / 2 + 1] +
                                             gamma * (1 - alpha) * et *
                                             weights[nts / 2 + 1:])
- 
+
                 # update l and b using equation-error formulation
                 l = l + phi * b + alpha * et
                 b = phi * b + alpha * beta * et
--- a/geomagio/algorithm/SqDistAlgorithm.py
+++ b/geomagio/algorithm/SqDistAlgorithm.py
@@ -436,6 +436,14 @@ class SqDistAlgorithm(Algorithm):
                 # performance.
                 r[i + 1] = gamma * (1 - alpha) * et / m
 
+                ## this properly smooths SQ locally; figure out if we need an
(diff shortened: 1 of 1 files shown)


aclaycomb-usgs:
Josh, the checker complained about 2 blank lines you had inserted in the file.  I know it's a pain, but can you remove those (between lines 350 and 380).

erigler-usgs:
It seemed to complain about a lot more than that. I had to insert spaces around all my arithmetic operators, which I honestly find not only tedious, but poor coding style. Anyway, I've made the changes, and pushed to origin, which seems to be keeping synced with this PR automatically. We'll see if Travis is pleased this time or not.

jmfee-usgs:
We adopted the coding standards used by the SciPy project with the goal of making our code easier to read/use for a broader audience:
https://docs.scipy.org/doc/scipy/reference/hacking.html 

You can run these checks locally (on every file save) if you install the utilities as described in the develop readme:
https://github.com/usgs/geomag-algorithms/blob/master/docs/develop.md

erigler-usgs:
That's not an explanation @jmfee-usgs , it's an excuse. As you should know, I know perfectly well where the standard comes from, and now it's wasted over half an hour of my day because an extra (and *intentional*) newline character or three wasn't "standard". This has nothing to do with readability, and everything to do with a process-over-product set of priorities.

PR #135, opened 2016-10-27
Issue#134

### LINKED ISSUE #134: Need clean Adjusted Phase 1 Matrix Generation Tool

aclaycomb-usgs:
adds generation tool, fixes #134


erigler-usgs:
@aclaycomb-usgs,

This looks mostly good. I have just a few minor comments and suggested changes I'd like you to address before I merge this pull request:

1. I consider the purpose of the Adjusted algorithm to transform heZ to XYZ. If you agree, you should change some of the notation in the Notebook. For example, the cells with the heading "**Calculate h,e,Z from H,D,Z**" should probably be "**Calculate X,Y,Z from H,D,Z**", and the variable names should be changed accordingly (e.g. *h_a* should be *X_a*).
2. Compare delta-F's on the same plot scale. 
3. After comparing delta-F's, you go on to compare/plot "**Raw h, Adjusted X, and ""delta-H""**". What you really want to compare is adjusted X with adjusted X, with the first being the adjusted X one would get with a simple static baseline adjustment, and the second being the adjusted X one would get with a static affine transformation. I understand that the first doesn't really have a name, since we've never actually done this operationally. Just explain what is going on in a sentence or two. And of course make similar changes for Y and Z.

Thanks,
Josh

erigler-usgs:
@aclaycomb-usgs,

1. Travis-CI is whining, but I can't imagine why since you only changed a Notebook.
2. Why did you comment out the lines that plot the deltas between the adjusted data done using just the static baselines, and the adjusted data using the affine transform?

aclaycomb-usgs:
I commented those out since they were not what you asked for in the previous comment, but you may want to look at the trend at some point.  I guess I could delete them since they're in a previous commit.

erigler-usgs:
Sorry, I wasn't clear. I think seeing the difference between the two types of "adjusted" data is useful. What I didn't find useful was the delta between the affine-adjusted XYZ and the raw HEZ, which is what you plotted in the original PR. I think you should uncomment those lines in the new one.

Do that, and figure out why Travis is whining, and I'll merge the PR.

aclaycomb-usgs:
Not sure why travis is failing, something must have changed on travis. Any ideas @jmfee-usgs @emartinez-usgs ?

erigler-usgs:
Everything looks good, except Travis-CI is failing for some reason related to SSL. Any idea what this is, or how to fix it @jmfee-usgs or @aclaycomb-usgs? I'm hesitant to "merge pull request" when Travis is upset.

jmfee-usgs:
Looks like it was a temporary issue, I restarted the travis build and it seems okay now.

PR #137, opened 2016-11-16
Minor fixes for local smoothing

erigler-usgs:
- updated init and process_one functions to handle "smooth" parameter
- included a *very* minor fix to updates in Sigma in additive function


erigler-usgs:
I discovered some minor bugs while using SqDistAlgorithm for analysis. Figured I should fix them before I completely forgot about them.

PR #138, opened 2016-11-28
Patch waveserver

jmfee-usgs:
Seeing intermittent memory leaks on EL7, when generating a large number of socket connections.  This is far from reproducible, but the client appears to hang while checking for data gaps and the only thing occurring at that point is accessing data using the earthworm client.  The code reading data from sockets is the primary suspect at this point, and this pull request provides a potential solution that could later be submitting to the upstream obspy code base:

- Add local copies of obspy earthworm client
- Add checks during loops so they will terminate if `socket.recv` (repeatedly) returns an empty string.

### DIFF
--- a/geomagio/edge/EdgeFactory.py
+++ b/geomagio/edge/EdgeFactory.py
@@ -15,12 +15,16 @@ import numpy
 import numpy.ma
 import obspy.core
 from datetime import datetime
-try:
-    # obspy 1.x
-    from obspy.clients import earthworm
-except:
-    # obspy 0.x
-    from obspy import earthworm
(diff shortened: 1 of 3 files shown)


jmfee-usgs:
From https://docs.python.org/2/howto/sockets.html#using-a-socket
> When a recv returns 0 bytes, it means the other side has closed (or is in the process of closing) the connection. You will not receive any more data on this connection. Ever.

jmfee-usgs:
A pull request for obspy has also been submitted:
https://github.com/obspy/obspy/pull/1605

PR #139, opened 2016-11-28
Update to MetaData.py based on recent measurements

ttwhite-USGS:
We recently reviewed our various data products and found discrepancy in some of the metadata.  We have updated these values based on recent measurements.  I've modified these values in ObservatoryMetadata.py to be consistent with our other data products.


jmfee-usgs:
See also https://github.com/usgs/geomag-edge-ws/pull/26

PR #140, opened 2016-12-15
Add make_cal.py script to generate CAL format from observation web se…

jmfee-usgs:
…rvice

@ttwhite-USGS : this script doesn't have any dependencies on this project, but this seemed like an appropriate place to store it for now.  The usage is slightly modified so now observatory and year are specified as arguments, so you don't need to edit the file.  Let me know what you think.


jmfee-usgs:
Bill and Jill have verified this script generates the correct output format.

PR #143, opened 2017-01-09
Pad to urlInterval in put_timeseries()

erigler-usgs:
TimeseriesFactory.get_timeseries() pads to the requested interval with NaNs,
filling in start/end gaps. TimeseriesFactory.put_timeseries() is supposed to
do this (according to Jeremy Fee), but it was trimming NaNs from the start/
end before writing to file. This PR fixes this by using ObsPy's Stream.trim()
method, in a similar manner to TimesereisFactory.get_timeseries(), to pad to
the specified start and end of the urlInterval..

### DIFF
--- a/geomagio/TimeseriesFactory.py
+++ b/geomagio/TimeseriesFactory.py
@@ -226,6 +226,13 @@ class TimeseriesFactory(object):
                         trace.stats.location = new_trace.stats.location
                     url_data = TimeseriesUtility.merge_streams(
                             existing_data, url_data)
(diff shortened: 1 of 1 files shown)


jmfee-usgs:
I'm thinking this should move after the existing file check and data merge, and to just before the file is written; so the same logic is applied for all puts whether the file exists or not:
https://github.com/usgs/geomag-algorithms/pull/143/files#diff-4a874991f91bb030399df14fe5936237R242

erigler-usgs:
I think I did what you were proposing @jmfee-usgs . Please confirm.

PR #149, opened 2017-01-21
Add Python 3 support

### LINKED ISSUE #145: Python 3 support
I'm interested in adding Python 3 support to this library. Are there any incompatibilities with Python 3? Is anyone else already working on this?

ysshah:
This pull request fixes #145 and adds support for Python 3.4 and 3.5. Changes are:
- Use absolute imports in all files
- Parentheses around print statements, and use `from __future__ import print_function` if writing to `sys.stderr` or printing multiple values
- Change `except` syntax from using comma to `as`
- Change `unicode` type to `str`
- Change `xrange` to `range`
- Import `StringIO` from `io`, not from `StringIO` or `cStringIO`
- Change dictionary keys function `data.keys()` to `list(data)` to get list of keys
- Change `/` to `//` whenever dividing integers

### DIFF
--- a/geomagio/Controller.py
+++ b/geomagio/Controller.py
@@ -1,5 +1,5 @@
 """Controller class for geomag algorithms"""
-from __future__ import absolute_import
+from __future__ import absolute_import, print_function
 
 import argparse
 import sys
@@ -207,8 +207,8 @@ class Controller(object):
         if options.update_limit != 0:
             if update_count >= options.update_limit:
                 return
-        print >> sys.stderr, 'checking gaps', \
-                options.starttime, options.endtime
+        print('checking gaps', options.starttime, options.endtime,
+            file=sys.stderr)
         algorithm = self._algorithm
         input_channels = options.inchannels or \
                 algorithm.get_input_channels()
@@ -254,8 +254,8 @@ class Controller(object):
             # fill gap
             options.starttime = output_gap[0]
             options.endtime = output_gap[1]
-            print >> sys.stderr, 'processing', \
-                    options.starttime, options.endtime
+            print('processing', options.starttime, options.endtime,
+                file=sys.stderr)
             self.run(options, input_timeseries)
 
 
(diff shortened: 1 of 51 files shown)


geomagio/iaga2002/IAGA2002Factory.py
```diff
@@ -57,7 +58,7 @@ def parse_string(self, data, observatory=None, interval='minute',
         endtime = obspy.core.UTCDateTime(parser.times[-1])
         data = parser.data
         stream = obspy.core.Stream()
-        length = len(data[data.keys()[0]])
+        length = len(data[list(data)[0]])
```
jmfee-usgs: Nice, this is simpler than the `list(data.keys())[0]` suggestion I've seen elsewhere.

ysshah: Thanks! I got the idea from [this handy website](http://python-future.org/compatible_idioms.html#dict-keys-values-items-as-a-list).

geomagio/TimeseriesUtility.py
```diff
@@ -47,7 +47,7 @@ def get_trace_gaps(trace):
     starttime = stats.starttime
     length = len(data)
     delta = stats.delta
-    for i in xrange(0, length):
+    for i in range(0, length):
```
ysshah: Yes, that's a good idea! I wasn't sure which functionality you wanted in Python 2. I've added that import line everywhere `xrange` was used.

geomagio/Controller.py
```diff
@@ -470,12 +470,12 @@ def main(args):
...
     if usingDeprecated:
-        print >> sys.stderr, 'WARNING: you are using deprecated arguments,' + \
-                ' please update your usage'
+        print('WARNING: you are using deprecated arguments,' +
+              ' please update your usage', file=sys.stderr)
     # TODO check for unused arguments.
 
     # make sure observatory is a tuple
-    if isinstance(args.observatory, (str, unicode)):
+    if isinstance(args.observatory, (str, str)):
```
ysshah: I think using `from builtins import str` would overwrite the `str` type everywhere in this file for Python 2. If you would like to preserve the distinction between `str` and `unicode` in Python 2, we can use
```python
from builtins import str as unicode
```
and change `str` back to `unicode` on this line. [This site](http://python-future.org/compatible_idioms.html#unicode) suggests importing it as `text`, but I believe it's just a matter of preference since the `unicode` variable isn't used anywhere else. I used this method in my following commit. What do you think?

PR #152, opened 2017-03-09
WIP: Refactor metadata

jmfee-usgs:
This is not yet fully integrated, and is a work in progress.  Creating pull request to get comments on the approach before I make any further changes.

Right now these files deal mainly with observatory level information, but with a little more structure could potentially also be used for SqDist and other configuration parameters; or those could be stored in a parallel metadata package.

I know we have ongoing discussions over storing metadata in a database or similar, but I would like to continue supporting offline processing using these or similar files.  If we integrate with another system, we could potentially add a script to download/update files like this.

### DIFF
--- a/geomagio/metadata/README.md
+++ b/geomagio/metadata/README.md
@@ -1,0 +1,80 @@
+# Metadata
+
+
+## Structure
+
+Each `type` of metadata uses a separate folder,
+with one file per `object` being described.
+
+```
+geomagio/metadata/TYPE/OBJECT.py
+```
+
+Each file defines a `METADATA` variable,
+which consists of a list of dictionary objects
+that define different versions of metadata associated with an `OBJECT`.
+
+
+```
+"""OBJECT metadata file"""
+
+METADATA = [
+    {VERSION2}
+    {VERSION1}
+]
+```
+
+
+### Epochs (Versions)
+
+Each version represents the metadata for a given time interval.
+*Versions must appear in descending order (newest versions first).*
+
+
+The time interval is represented using the reserved properties
+`starttime` and `endtime`, which define the time interval that this version of
+metadata are valid.
+
+The interval can be represented using set notation as:
+    `[starttime, endtime)`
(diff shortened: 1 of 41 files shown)


erigler-usgs:
@jmfee-usgs:

Just a few initial thoughts on the proposed metadata package:

- Is there a compelling reason to implement these as .py files instead of, for example, JSON?

- multi-observatory derived products (e.g., Dst) should be a primary consideration. At the least, we need some way to track different input sources.

- A lower priority, but still important for long-term plans, gridded data sets should be considered. At the least, we need to be able to store coordinates as a list.

jmfee-usgs:
@erigler-usgs :

- We could consider a different format like JSON.  In either case, the data is being read for use within a python script.  My main concern is this be able to work without a network connection, if data is already stored locally.

- This is currently focused on observatory metadata, one of the reasons these are in an `observatory` directory.  I think we should store product/element/channel metadata separately.

- I think gridded data fits in with the product/element/channel metadata, but I'm not familiar enough with the metadata needs for those data sets.  Would they reference observatory, element, and algorithm metadata as part of the data files?

erigler-usgs:
- If meta data is essentially static, although with a possibility for versions, then a JSON format seems better if only because it could be read in by different software. But I can imagine a scenario where you might want dynamic metadata, like magnetic coordinates that evolve slowly with the Earth's changing magnetic field...but that should probably be a product, not metadata.

- OK, I understand your response to bullet 2.

- Gridded data, or more accurately, data registered to a fixed set of coordinates that are synchronous in time, could reference observatory, element, or algorithm metadata. Ideally, I want to be able to specify a grid of "virtual observatories" that holds a full vector complement (e.g., XYZF), calculated from a given algorithm. I know that's a lot to consider, and is why I said it's a lower priority for short term plans. But I also know you like to plan ahead, so I threw it out there.

jmfee-usgs:
this is still needed, but is out of date

PR #167, opened 2017-08-03
Web Service POC

hschovanec-usgs:
This includes additions to parse query strings for the geomag web service. It also includes data fetching using the Edge Factory based on user input and defaults. There needs to be the addition of error output and more unit testing. Would appreciate comments on this work in progress before proceeding with more additions. jmfee-usgs/geomag-algorithms#1

### DIFF
--- a/geomagio/WebService.py
+++ b/geomagio/WebService.py
@@ -4,6 +4,8 @@
 from __future__ import print_function
 from cgi import parse_qs, escape
 from datetime import datetime
+from json import load
+import os.path
 import sys
 
 from geomagio.edge import EdgeFactory
@@ -16,6 +18,15 @@ DEFAULT_DATA_TYPE = 'variation'
 DEFAULT_ELEMENTS = ('X', 'Y', 'Z', 'F')
 DEFAULT_OUTPUT_FORMAT = 'iaga2002'
 DEFAULT_SAMPLING_PERIOD = '60'
+ERROR_CODE_MESSAGES = {
+        204: 'No Data',
+        400: 'Bad Request',
+        404: 'Not Found',
+        409: 'Conflict',
+        500: 'Internal Server Error',
+        501: 'Not Implemented',
+        503: 'Service Unavailable'
+}
 VALID_DATA_TYPES = [
         'variation',
         'adjusted',
@@ -45,12 +56,12 @@ def _get_param(params, key, required=False):
 
     Raises
     ------
-    WebServiceError
+    WebServiceException
         if the parameter is specified more than once
         or if required paramenter is not specified.
     """
     value = params.get(key)
-    if type(value) in (list, tuple):
+    if isinstance(value, (list, tuple)):
         if len(value) > 1:
             raise WebServiceException('"' + key +
                     '" may only be specified once.')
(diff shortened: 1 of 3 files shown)


jmfee-usgs:
See comments on https://github.com/jmfee-usgs/geomag-algorithms/pull/1

geomagio/WebService.py
```diff
@@ -0,0 +1,224 @@
...
+        # fetch data
+        data = self.fetch(query)
+        # format data
+        data_string = self.format(query, data)
+        # send response
+        start_response('200 OK',
+                [
+                    ("Content-Type", "text/plain")
+                ])
+        if isinstance(data_string, str):
```
jmfee-usgs: shift this encoding step before start_response, since an error may occur during encoding.

geomagio/WebService.py
```diff
@@ -0,0 +1,224 @@
...
+        query = {}
+        query = WebServiceQuery.parse(environ['QUERY_STRING'])
+        # fetch data
+        data = self.fetch(query)
+        # format data
+        data_string = self.format(query, data)
+        # send response
+        start_response('200 OK',
+                [
+                    ("Content-Type", "text/plain")
```
jmfee-usgs: The mime type varies based on output format, so this may need to move into an output/success (the opposite of error) method.

geomagio/WebService.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+class WebService(object):
+    def __init__(self, factory):
+        self.factory = factory
+
+    def __call__(self, environ, start_response):
+        """Implement WSGI interface"""
+        # parse params
+        query = {}
```
jmfee-usgs: this line is not needed

geomagio/WebService.py
```diff
@@ -0,0 +1,224 @@
...
+        elements = escape(elements)
+        sampling_period = escape(sampling_period)
+        type = escape(type)
+        format = escape(format)
+        # Check for parameters and set defaults
+        if not id:
+            raise WebServiceException(
+                'Missing observatory id.')
+        now = datetime.now()
+        if starttime and endtime:
```
jmfee-usgs: validate starttime and endtime separately:

- if starttime is not set, default to the start of the current day
- if endtime is not set, default to starttime plus 24 hours
- if starttime or endtime are invalid, validation error
- if endtime is before starttime, validation error

geomagio/WebService.py
```diff
@@ -0,0 +1,224 @@
...
+        Raises
+        ------
+        TimeseriesFactoryException
+            if id, type, sampling_period, or format are not supported.
+        """
+        # Create dictionary of lists
+        dict = parse_qs(params)
+        # Return values
+        id = dict.get('id', [''])[0]
+        starttime = dict.get('starttime', [''])[0]
```
jmfee-usgs: if parameters (besides elements) have multiple values (the array has more than one element) it should probably be considered a validation error

geomagio/WebService.py
```diff
@@ -0,0 +1,224 @@
...
+            endtime = starttime  + (24 * 60 * 60 - 1)
+        if starttime and not endtime:
+            starttime = UTCDateTime(starttime)
+            endtime = starttime  + (24 * 60 * 60 - 1)
+        if not starttime and endtime:
+            raise WebServiceException(
+                    'Missing start time.')
+        if elements:
+            elements = [el.strip().upper() for el in elements.split(',')]
+        if not elements:
```
jmfee-usgs: Make this an else, so the boolean condition is only evaluated once.

geomagio/WebService.py
```diff
@@ -0,0 +1,224 @@
...
+            elements = ('X', 'Y', 'Z', 'F')
+        valid_periods = ['1', '60']
```
jmfee-usgs: make constants globals at the top of this module.

geomagio/WebService.py
```diff
@@ -0,0 +1,224 @@
...
+        if sampling_period not in valid_periods:
+            raise WebServiceException(
+                    'Invalid sampling period.'\
+                    ' Valid sampling periods: %s' % valid_periods)
+        # TODO: Add hourly option
+        if sampling_period == '1':
+            sampling_period = 'second'
+        if sampling_period == '60':
+            sampling_period = 'minute'
+        valid_types = [
```
jmfee-usgs: make constants globals at the top of the module

geomagio/WebService.py
```diff
@@ -0,0 +1,224 @@
...
+                ' Valid data types: %s' % valid_types)
+        # TODO: Add json to valid formats
+        valid_formats = ['iaga2002']
+        if not format:
+            format = 'iaga2002'
+        if format not in valid_formats:
+            raise WebServiceException(
+                'Invalid format.'\
+                ' Valid formats: %s' % valid_formats)
+        # Fill dictionary with parameters and return
```
jmfee-usgs: Create a WebServiceQuery object then set query properties, instead of using a dictionary

geomagio/WebService.py
```diff
@@ -0,0 +1,224 @@
...
+        if not id:
+            raise WebServiceException(
+                'Missing observatory id.')
+        now = datetime.now()
+        if starttime and endtime:
+            starttime = UTCDateTime(starttime)
+            endtime = UTCDateTime(endtime)
+        if not starttime and not endtime:
+            starttime = UTCDateTime(
+                        year=now.year,
```
jmfee-usgs: When continuing on multiple lines, use 2 indentations (8 spaces total)

geomagio/WebService.py
```diff
@@ -126,88 +138,116 @@ def parse(cls, params):
...
-        # Return values
-        id = dict.get('id', [''])[0]
-        starttime = dict.get('starttime', [''])[0]
-        endtime = dict.get('endtime', [''])[0]
-        elements = dict.get('elements', [''])[0]
-        sampling_period = dict.get('sampling_period', [''])[0]
-        type = dict.get('type', [''])[0]
-        format = dict.get('format', [''])[0]
+        # Get values
+        if len(dict.get('id', [])) <= 1:
```
jmfee-usgs: This appears to only check if there are 0 values, rather than more than 1.  I recommend a separate method for this repetitive validation.  Something like:

```
def get_param(params, key, default=None, required=False):
    """Get parameter from dictionary.

    Parameters
    ----------
    params : dict
        parameters dictionary.
    key : str
        parameter name
    default : any
        default value if key not set in dictionary
    required : bool
        whether parameter is required.

    Returns
    -------
    value : str
        value from dictionary, or value of default parameter if value not found in dictionary.
    """
    value = dict.get(key)
    if type(value) in (list, tuple):
        if len(value) > 1:
            raise WebServiceException('"' + key + '" may only be specified once')
        value = value[0]
    if value is None:
        if required and not default:
            raise WebServiceException('"' + key + '" is a required parameter')
        return default
    return value
```

jmfee-usgs:
- make verify parameters a method on the WebServiceQuery object (it uses web service query attributes)

- move error formatting to a WebService object method
  - explicitly pass code (400)
  - make code => code message dictionary a global
  - remove WebServiceError class, only use WebServiceException

- don't read version from SERVER_SOFTWARE; make version a parameter to WebService class similar to (possibly read from setup.py or package.json when starting service)

- in __call__:
  - parse and validate the query object in one try block; use 400 Bad Request if there are errors
  - fetch and format data in another try block; use 500 Internal Server Error if there are errors
  - move start_response into error/format methods on the WebService class

- in fetch, don't update the query.sampling_period value; use a local variable instead

Issue #164, opened 2017-08-01
Add json output factory


jmfee-usgs:
Similar to 

https://github.com/usgs/geomag-edge-ws/blob/master/src/lib/classes/JsonOutputFormat.class.php

PR #168, opened 2017-08-03
Add json output factory

hschovanec-usgs:
This is a work in progress to aid with the web service POC #167 and #164. It has been tested using the function call format in WebService.py, data = JSONWriter.format(data, ('H', 'E', 'Z', 'F')), and in a Jupyter notebook for outputting files. jmfee-usgs/geomag-algorithms#2

### DIFF
--- a/geomagio/imfjson/IMFJSONWriter.py
+++ b/geomagio/imfjson/IMFJSONWriter.py
@@ -88,9 +88,11 @@ class IMFJSONWriter(object):
             # TODO: Add flag metadata
             values += [value_dict]
             series = np.copy(trace.data)
-            series = ChannelConverter.get_minutes_from_radians(series)
+            if c == 'D':
+                series = ChannelConverter.get_minutes_from_radians(series)
             series[np.isnan(series)] = None
-            value_dict['values'] = series
+            # Converting numpy array to list required for JSON serialization
+            value_dict['values'] = series.tolist()
         return values
 
     def _format_metadata(self, stats, channels):
--- a/geomagio/imfjson/IMFJSONWriter.py
+++ b/geomagio/imfjson/IMFJSONWriter.py
@@ -8,6 +8,7 @@ import numpy as np
 from .. import ChannelConverter, TimeseriesUtility
 from ..TimeseriesFactoryException import TimeseriesFactoryException
 
+
 class IMFJSONWriter(object):
     """JSON writer.
     """
@@ -67,9 +68,6 @@ class IMFJSONWriter(object):
         array_like
             an array containing dictionaries of data.
         """
-        if timeseries.select(channel='D'):
-            d = timeseries.select(channel='D')
-            d[0].data = ChannelConverter.get_minutes_from_radians(d[0].data)
         values = []
         for c in channels:
             value_dict = OrderedDict()
@@ -89,9 +87,10 @@ class IMFJSONWriter(object):
                 value_dict['location'] = stats.location
             # TODO: Add flag metadata
             values += [value_dict]
-            data = np.copy(trace.data)
-            data[np.isnan(data)] = None
-            value_dict['values'] = data.tolist()
+            series = np.copy(trace.data)
+            series = ChannelConverter.get_minutes_from_radians(series)
+            series[np.isnan(series)] = None
+            value_dict['values'] = series
         return values
 
     def _format_metadata(self, stats, channels):
--- a/geomagio/imfjson/IMFJSONWriter.py
+++ b/geomagio/imfjson/IMFJSONWriter.py
@@ -89,9 +89,10 @@ class IMFJSONWriter(object):
             series = np.copy(trace.data)
             if c == 'D':
                 series = ChannelConverter.get_minutes_from_radians(series)
-            series[np.isnan(series)] = None
             # Converting numpy array to list required for JSON serialization
-            value_dict['values'] = series.tolist()
+            series = series.tolist()
+            series = [None if str(x) == 'nan' else x for x in series]
+            value_dict['values'] = series
             # TODO: Add flag metadata
         return values
 
--- a/geomagio/imfjson/IMFJSONWriter.py
+++ b/geomagio/imfjson/IMFJSONWriter.py
@@ -1,18 +1,15 @@
 from __future__ import absolute_import
-from builtins import range
 
 from io import BytesIO
 from collections import OrderedDict
 from datetime import datetime
 import json
 import numpy as np
-import textwrap
 from .. import ChannelConverter, TimeseriesUtility
 from ..edge import EdgeFactory
 from ..TimeseriesFactoryException import TimeseriesFactoryException
-from ..Util import create_empty_trace
 
-class JSONWriter(object):
+class IMFJSONWriter(object):
     """JSON writer.
     """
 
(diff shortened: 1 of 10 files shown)


jmfee-usgs:
See comments on https://github.com/jmfee-usgs/geomag-algorithms/pull/2

geomagio/imfjson/IMFJSONWriter.py
```diff
@@ -0,0 +1,217 @@
...
+            else:
+                edge_channel = c
+            value_dict['channel'] = edge_channel
+            if 'location' in stats:
+                value_dict['location'] = stats.location
+            # TODO: Add flag metadata
+            values += [value_dict]
+            data = np.copy(trace.data)
+            data[np.isnan(data)] = None
+            value_dict['values'] = data.tolist()
```
jmfee-usgs: does this need to be converted to a list, or will it work with the numpy array?

geomagio/imfjson/IMFJSONWriter.py
```diff
@@ -0,0 +1,217 @@
...
+            holds the observatory metadata
+
+        Returns
+        -------
+        array_like
+            an array containing dictionaries of data.
+        """
+        if timeseries.select(channel='D'):
+            d = timeseries.select(channel='D')
+            d[0].data = ChannelConverter.get_minutes_from_radians(d[0].data)
```
jmfee-usgs: Do this after making a copy of the trace data (on line 92), so the original trace remains unchanged.

```
data = np.copy(trace.data)
if c == 'D':
  data = ChannelConverter.get_minutes_from_radians(data)
...
```

geomagio/imfjson/IMFJSONWriter.py
hschovanec-usgs: The numpy array was converted to a list to, because numpy arrays are not JSON serializable. However, if altered data type and dimension is of concern it should be possible to convert the array using pickle or base64 instead.

jmfee-usgs: I wasn't sure if it was needed.  Add a comment to that line that it's require for JSON serialization.

geomagio/imfjson/IMFJSONWriter.py
```diff
@@ -0,0 +1,218 @@
...
+        request = kwargs.get('request')
+        for channel in channels:
+            if timeseries.select(channel=channel).count() == 0:
+                raise TimeseriesFactoryException(
+                    'Missing channel "%s" for output, available channels %s' %
+                    (channel, str(TimeseriesUtility.get_channels(timeseries))))
+        stats = timeseries[0].stats
+        file_dict['type'] = 'Timeseries'
+        file_dict['metadata'] = self._format_metadata(stats, channels)
+        if request:
```
jmfee-usgs: Move this logic out of the writer, accept a `url` parameter and include it in the metadata regardless of value.  The web service can figure out the url.

geomagio/imfjson/IMFJSONWriter.py
```diff
@@ -0,0 +1,218 @@
...
+        array_like
+            an array containing dictionaries of data.
+        """
+        values = []
+        for c in channels:
+            value_dict = OrderedDict()
+            trace = timeseries.select(channel=c)[0]
+            value_dict['id'] = c
+            value_dict = OrderedDict()
+            value_dict['element'] = c
```
jmfee-usgs: These parameters (besides `id`) should go under a `metadata` object.  See the output from the php service:
https://geomag.usgs.gov/ws/edge/?starttime=2018-01-02&endtime=2018-01-02T00:10:00&id=BOU&elements=H,E,Z&format=json

geomagio/imfjson/IMFJSONWriter.py
```diff
@@ -0,0 +1,218 @@
...
+            if 'network' in stats:
```
jmfee-usgs: I think `network`, `channel`, and `location` are always defined, even if they default to empty string.

geomagio/imfjson/IMFJSONWriter.py
```diff
@@ -0,0 +1,218 @@
...
+            value_dict['station'] = stats.station
+            if 'channel' in stats:
+                edge_channel = trace.stats.channel
+            else:
+                edge_channel = c
+            value_dict['channel'] = edge_channel
+            if 'location' in stats:
+                value_dict['location'] = stats.location
+            # TODO: Add flag metadata
+            values += [value_dict]
```
jmfee-usgs: put this at the end, after `values` are added

geomagio/imfjson/IMFJSONWriter.py
```diff
@@ -0,0 +1,218 @@
...
+        """
+        metadata_dict = OrderedDict()
+        intermag = OrderedDict()
+        imo = OrderedDict()
+        imo['iaga_code'] = stats.station
+        if 'station_name' in stats:
+            imo['name'] = stats.station_name
+        coords = [None] * 3
+        if 'geodetic_longitude' in stats:
+            coords[0] = str(stats.geodetic_longitude)
```
jmfee-usgs: numeric values should still be numeric in json output

geomagio/imfjson/IMFJSONWriter.py
```diff
@@ -0,0 +1,218 @@
...
+                rate = 60
+            elif stats.sampling_rate == 1. / 3600.:
+                rate = 3600
+            elif stats.sampling_rate == 1. / 86400.:
+                rate = 86400
+            else:
+                rate = 1
+            intermag['sampling_period'] = str(rate)
+        if 'sensor_sampling_rate' in stats:
+            sampling = 1 / stats.sensor_sampling_rate
```
jmfee-usgs: I think this should be the rate, instead of 1 / rate

geomagio/imfjson/IMFJSONWriter.py
```diff
@@ -0,0 +1,218 @@
...
+                    (channel, str(TimeseriesUtility.get_channels(timeseries))))
+        stats = timeseries[0].stats
+        file_dict['type'] = 'Timeseries'
+        file_dict['metadata'] = self._format_metadata(stats, channels)
+        if request:
+            base = 'http://geomag.usgs.gov/ws/edge/?'
+            file_dict['metadata']['url'] = base + request
+        file_dict['times'] = self._format_times(timeseries, channels)
+        file_dict['values'] = self._format_data(timeseries, channels, stats)
+        formatted_timeseries = json.dumps(file_dict,
```
jmfee-usgs: Use `separators=(',', ':')` to create smaller json output

test/imfjson_test/IMFJSONWriter_test.py
```diff
@@ -0,0 +1,112 @@
+"""Tests for the IMFJSON Writer class."""
+
+from nose.tools import assert_equals
+from geomagio import edge
+from geomagio.imfjson import IMFJSONWriter
+
+from obspy.core import UTCDateTime
+
+EXAMPLE_INPUT_FACTORY = edge.EdgeFactory()
```
jmfee-usgs: Use a local factory so an internet connection is not required to run tests...  One option is an iaga2002 factory using files in `etc/iaga2002`

PR #177, opened 2017-09-07
Added unit tests for the web service

hschovanec-usgs:
Added unit tests for the web service.

Code Review Process Checklist
- [ ] Will the pull request merge into the correct branch?
- [ ] Pull the branch
  - [ ] Does the tests run correctly?
- [ ] Go through the changes:
  - [ ] Does it conform to the architecture?
  - [ ] Are they inline with the coding standards?
  - [ ] Do they make sense?
  - [ ] Are they clean?
  - [ ] Easily readable?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ before_install:
   - conda info -a
 install:
   - conda config --add channels conda-forge
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION obspy pycurl nose flake8 coverage
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION obspy pycurl nose flake8 coverage webtest
   - source activate test-environment
 script:
   - flake8 --config=.flake8 bin/ geomagio/ test/
--- a/geomagio/WebService.py
+++ b/geomagio/WebService.py
@@ -107,7 +107,7 @@ class WebService(object):
                 timeseries_string = timeseries_string.encode('utf8')
         except Exception:
             exception = sys.exc_info()[1]
-            message = exception.args[0]
+            message = "Server error."
             error_body = self.error(500, message, environ, start_response)
             return [error_body]
         return [timeseries_string]
--- a/geomagio/WebService.py
+++ b/geomagio/WebService.py
@@ -6,7 +6,6 @@ from cgi import escape, parse_qs
 from collections import OrderedDict
 from datetime import datetime
 from json import dumps
-import sys
 
 from geomagio.edge import EdgeFactory
 from geomagio.iaga2002 import IAGA2002Writer
@@ -87,9 +86,8 @@ class WebService(object):
             query = self.parse(parse_qs(environ['QUERY_STRING']))
             query._verify_parameters()
             self.output_format = query.output_format
-        except Exception:
-            exception = sys.exc_info()[1]
-            message = exception.args[0]
+        except Exception as e:
+            message = str(e)
             ftype = parse_qs(environ['QUERY_STRING']).get('format', [''])[0]
             if ftype == 'json':
                 self.output_format = 'json'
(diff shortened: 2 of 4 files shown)


test/WebService_test.py
```diff
@@ -0,0 +1,155 @@
+"""Unit Tests for WebService"""
+from cgi import parse_qs
+from datetime import datetime
+from nose.tools import assert_equals
```
jmfee-usgs: combine imports from the same package, e.g.

```
from nose.tools import assert_equals, assert_raises, ...
```

test/WebService_test.py
```diff
@@ -0,0 +1,155 @@
...
+from webtest import TestApp
+
+from geomagio.edge import EdgeFactory
+from geomagio.WebService import _get_param
+from geomagio.WebService import WebService
+from obspy.core.stream import Stream
+from obspy.core.utcdatetime import UTCDateTime
+
+
+APP = TestApp(WebService(EdgeFactory()))
```
jmfee-usgs: Use a factory that doesn't require an internet connection.  `WebService` only uses the `get_timeseries` method, so you can create a test factory that returns a test `Timeseries` object

.travis.yml
```diff
@@ -21,7 +21,7 @@ before_install:
   - conda info -a
 install:
   - conda config --add channels conda-forge
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION obspy pycurl nose flake8 coverage
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION obspy pycurl nose flake8 coverage webtest
```
jmfee-usgs: If adding a development/testing dependency, also update docs/develop.md

test/WebService_test.py
```diff
@@ -0,0 +1,155 @@
...
+            '/?id=BOU&format=IAGA2002'
+    ]
+    for request in valid_requests:
+        try:
+            response = APP.get(request)
+            assert_equals(response.status_int, 200)
+            assert_equals(response.status, '200 OK')
+            assert_equals(response.content_type, 'text/plain')
+        except Exception:
+            exception = sys.exc_info()[1]
```
jmfee-usgs: add a variable to the except clause, instead of using sys.exc_info
```
except Exception as e:
```

test/WebService_test.py
```diff
@@ -0,0 +1,155 @@
...
+    assert_equals(query.data_type, 'variation')
+
+
+def test_requests():
+    """WebService_test.test_requests()
+
+    Use TestApp to confirm correct response status, status int,
+    and content-type.
+    """
+    # Check invalid request (bad values)
```
jmfee-usgs: The `parse` test can handle most of these cases.  For the requests, make sure that a parse error generates the 400 Bad Request response, a server error generates a 500 Internal Server Error response, and success generates a 200 OK response.

geomagio/WebService.py
```diff
@@ -98,7 +98,7 @@ def __call__(self, environ, start_response):
                 timeseries_string = timeseries_string.encode('utf8')
         except Exception:
```
jmfee-usgs: This can use the exception message, but shouldn't use `sys.exc_info` to get the exception...

```
except Exception as e:
    message = str(e)
```

test/WebService_test.py
```diff
@@ -0,0 +1,146 @@
...
+    assert_raises(Exception, _get_param, params, 'sampling_period')
+
+
+def test_fetch():
+    """WebService_test.test_fetch())
+
+    Call function WebService.fetch to confirm tht it returns an
+    obspy.core.stream object.
+    """
+    service = WebService(EdgeFactory())
```
jmfee-usgs: this test still requires the internet, use the test factory.

PR #183, opened 2017-10-02
Determine interval from delta when data is not from EDGE.

jmfee-usgs:
Add optional --output-observatory argument to controller to support average algorithm.


jmfee-usgs:
already merged

PR #185, opened 2017-11-01
Better estimation of data_time in imfv283

jmfee-usgs:
NRCan GPS times are incorrect.  This pull request handles the existing error, as well as the upcoming errors in 2019...

### DIFF
--- a/geomagio/imfv283/IMFV283Parser.py
+++ b/geomagio/imfv283/IMFV283Parser.py
@@ -2,12 +2,12 @@
 from __future__ import absolute_import
 from builtins import range
 
-import math
 import numpy
 import sys
 import obspy
-from obspy.core import UTCDateTime
 
+from datetime import datetime, timedelta
+from obspy import UTCDateTime
 from . import imfv283_codes
 
 # values that represent missing data points in IAGA2002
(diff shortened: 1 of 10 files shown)


geomagio/imfv283/IMFV283Parser.py
```diff
@@ -99,6 +99,73 @@ def parse(self, data):
...
+        data_time : UTCDateTime
+            estimated time of first data sample
+        transmit_time : UTCDateTime
+            time when message was transmitted
+        corrected : bool
+            whether a gps timing error was corrected
+        """
+        # convert to datetime
+        transmit_time = UTCDateTime(
+                '20' + transmission[0:5] + 'T' + transmission[5:])
```
erigler-usgs: Any chance this would be applied retrospectively to data from the 20th century? Maybe more to the point, aren't there standard libraries that, however imperfectly, implement a standard procedure for dealing with 2-digit years?

geomagio/imfv283/IMFV283Parser.py
```diff
@@ -99,6 +99,73 @@ def parse(self, data):
...
+        transmit_year = transmit_time.year
+        # delta should not include first day of year
+        data_time_delta = timedelta(days=doy - 1, minutes=minute)
+        data_time = UTCDateTime(
+                datetime(transmit_year, 1, 1) + data_time_delta)
+        if data_time > transmit_time:
+            # data cannot be measured after it is transmitted
```
erigler-usgs: This comment is technically correct, but that's not really the problem this block is addressing. The comment should reflect the specific issue being addressed. Namely, when the data is transmitted after the new year roll-over, but the actual data time stamps are in the previous year. And again, will this function ever be applied retrospectively, where it might happen that  `transmit_year` is more than 1 year ahead of the data time stamp?

geomagio/imfv283/IMFV283Parser.py
jmfee-usgs: That is not currently planned.  The previous code added the integer value of transmission[0:2] to 2000 to determine the transmission year.  If older values are used, pre-2000, it will generate errors elsewhere in the code.

geomagio/imfv283/IMFV283Parser.py
jmfee-usgs: The parser considers it an error if the data is not within 24 minutes of transmission time.

geomagio/imfv283/IMFV283Parser.py
```diff
@@ -278,9 +304,17 @@ def _post_process(self, data, msg_header, goes_header):
...
+                goes_header['day'],
+                goes_header['minute'])
+        if corrected:
+            sys.stderr.write(
+                'corrected gps week number error\n' +
+                '\ttransmit day={}, reported day={}, corrected day={}\n'
+                .format(msg_time.julday, goes_header['day'], goes_time.julday))
         if (msg_time - goes_time) > (24 * 60):
-            sys.stderr.write('data over twice as old as the message')
+            sys.stderr.write('data over twice as old as the message\n')
```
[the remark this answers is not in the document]
jmfee-usgs: Agreed, we should have better logic about this.  However, I believe it was chosen because 12 minute blocks are transmitted with an up to 12 minute delay for the goes transmission window.

PR #209, opened 2018-07-19
Required Channels Implementation for XYZ Algorithm

arigdon-usgs:
Generally only two channels are required for converting in the XYZ Algorithm as Z and F stay the same. While Barrow's F channel was down the algorithm would fail saying that the F channel was necessary when it could be done without the F channel. This update implements a get_required_channels() function in the Algorithm so that get_stream_gaps() and can_produce_data() will only check for the required channels. Changes in the StreamConverter updated how the Z and F channels were being used so that empty or non-existent Z and F channels could be passed without causing the algorithm to fail.

### DIFF
--- a/geomagio/algorithm/Algorithm.py
+++ b/geomagio/algorithm/Algorithm.py
@@ -87,7 +87,7 @@ class Algorithm(object):
         """
         return (start, end)
 
-    def can_produce_data(self, starttime, endtime, stream, channels = None):
+    def can_produce_data(self, starttime, endtime, stream, channels=None):
         """Can Product data
 
         Parameters
@@ -100,7 +100,7 @@ class Algorithm(object):
             The input stream we want to make certain has data for the algorithm
         """
         input_gaps = TimeseriesUtility.get_merged_gaps(
-                TimeseriesUtility.get_stream_gaps(stream,channels))
+                TimeseriesUtility.get_stream_gaps(stream, channels))
         for input_gap in input_gaps:
             # Check for gaps that include the entire range
             if (starttime >= input_gap[0] and
--- a/geomagio/algorithm/Algorithm.py
+++ b/geomagio/algorithm/Algorithm.py
@@ -87,7 +87,7 @@ class Algorithm(object):
         """
         return (start, end)
 
-    def can_produce_data(self, starttime, endtime, stream):
+    def can_produce_data(self, starttime, endtime, stream, channels = None):
         """Can Product data
 
         Parameters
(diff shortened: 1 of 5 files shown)


geomagio/algorithm/Algorithm.py
```diff
@@ -77,7 +87,7 @@ def get_input_interval(self, start, end, observatory=None, channels=None):
         """
         return (start, end)
 
-    def can_produce_data(self, starttime, endtime, stream):
+    def can_produce_data(self, starttime, endtime, stream, channels=None):
```
jmfee-usgs: is this parameter required?  Can you just call get_stream_gaps() with get_required_channels()?

geomagio/algorithm/XYZAlgorithm.py
```diff
@@ -134,5 +140,15 @@ def configure(self, arguments):
         """
         self._informat = arguments.xyz_from
         self._outformat = arguments.xyz_to
-        self._inchannels = CHANNELS[self._informat]
-        self._outchannels = CHANNELS[self._outformat]
+        self._inchannels = arguments.inchannels or \
```
jmfee-usgs: Is this still required?  I think we still want these to the the format channels, and then only check for gaps on the specified output channels here?
https://github.com/usgs/geomag-algorithms/pull/209/files#diff-eb4ccf738235c6cfed6c4639b24600daR231

jmfee-usgs:
I'm not too worried about the codacy check, you're following the api; and fixing that is not part of this issue.  Please take a look at the other comments, happy to discuss

geomagio/algorithm/Algorithm.py
arigdon-usgs: In the Algorithm.py the get_required_stream() returns None for anything other than XYZAlgorithm. It looks like the get_stream_gaps() will check every trace if the channels parameter is None. If that is the case then yes I think we can drop the channels parameter in the can_produce_data().

erigler-usgs: I'm not entirely sure I understand @arigdon-usgs 's response, so perhaps he can clarify, but without the `channels=None` parameter here, the change made in the Controller's call to can_produce_data() will not work.

jmfee-usgs: @erigler-usgs : since get_required_stream() returns None (as @arigdon-usgs points out), I think this will work; basically have `can_produce_data` call `get_required_channels` directly, instead of making algorithms call get_required_data manually and pass to can_produce_data.

erigler-usgs: To be clear, he'd need to revert the change made to the Controller (i.e., https://github.com/arigdon-usgs/geomag-algorithms/blob/b719be106acb8a5002c7de5a1dbd443f205f597c/geomagio/Controller.py#L249), correct?

arigdon-usgs: @erigler : Yes I did have to go and revert the changes in the Controller.

jmfee-usgs:
@arigdon-usgs : Please add a unit tests to verify
- XYZ can process only H,E and produce D
- XYZ can process and produce D when H,E,Z,F are provided but F has gaps

PR #214, opened 2018-07-27
Merge streams fix

arigdon-usgs:
Do I need to rebase so that the travis.yml and EdgeFactory fixes do not show up? Either way this fixes an issue where the merge streams completely drops a channel if it is empty. This just saves that channel, trims it to the new start and end times, and then adds the channel back to the stream before returning the full stream. Created a unit test for this as well. One issue is it seams the endtimes after the merge streams seem to be at 59 seconds instead of on the minute which is expected for minute data.

### DIFF
--- a/geomagio/TimeseriesUtility.py
+++ b/geomagio/TimeseriesUtility.py
@@ -236,15 +236,16 @@ def merge_streams(*streams):
 =======
 =======
 
+<<<<<<< HEAD
 >>>>>>> fixed rebasing issues
     # print('\n masked trace \n')
     # print(masked_trace)
     # trim masked trace to the same size as other traces and add back to merged stream
+=======
+>>>>>>> Fixed linting and spacing errors and cleaned up code
     if masked_trace:
-        masked_trace.trim(merged[0].stats.starttime,merged[0].stats.endtime)
+        masked_trace.trim(merged[0].stats.starttime, merged[0].stats.endtime)
         merged += masked_trace
-    # print('\n after merging masked trace back in \n')
-    # print(merged)
 
 <<<<<<< HEAD
 
--- a/geomagio/TimeseriesUtility.py
+++ b/geomagio/TimeseriesUtility.py
@@ -195,35 +195,30 @@ def merge_streams(*streams):
         stream with contiguous traces merged, and gaps filled with numpy.nan
     """
     merged = obspy.core.Stream()
-    masked_trace = None
+    masked = obspy.core.Stream()
 
-    # add unmasked, split traces to be merged
+    # sort out empty
     for stream in streams:
-        merged += mask_stream(stream)
-
-    # if trace is completely masked separate out to be added back in later
-    for trace in merged:
-        if trace.data.mask.all():
-            if not masked_trace:
-                masked_trace = trace
+        for trace in stream:
+            if numpy.isnan(trace.data).all():
+                masked += trace
             else:
-                masked_trace += trace
+                merged += trace
+
+    merged = mask_stream(merged)
 
     # split traces that contain gaps
     merged = merged.split()
+
+    merged += masked
 
     # merge data
     merged.merge(
             # 1 = do not interpolate
             interpolation_samples=1,
+            fill_value=numpy.NaN,
             # 1 = when there is overlap, use data from trace with last endtime
             method=1)
-
-    # trim masked trace to the same size as other traces (not done in merge())
-    # and add back to merged stream
-    if masked_trace:
-        masked_trace.trim(merged[0].stats.starttime, merged[0].stats.endtime)
-        merged += masked_trace
 
     # convert back to NaN filled array
     merged = unmask_stream(merged)
--- a/geomagio/TimeseriesUtility.py
+++ b/geomagio/TimeseriesUtility.py
@@ -195,9 +195,26 @@ def merge_streams(*streams):
         stream with contiguous traces merged, and gaps filled with numpy.nan
     """
     merged = obspy.core.Stream()
+<<<<<<< HEAD
     # add unmasked, split traces to be merged
     for stream in streams:
         merged += mask_stream(stream)
+=======
+    masked_trace = None
+
+    # add unmasked, split traces to be merged
+    for stream in streams:
+        merged += mask_stream(stream)
+
+    # if trace is completely masked separate out to be added back in later
+    for trace in merged:
+        if trace.data.mask.all():
+            if not masked_trace:
+                masked_trace = trace
+            else:
+                masked_trace += trace
+
+>>>>>>> Merge stream issues with empty channels
     # split traces that contain gaps
     merged = merged.split()
     # merge data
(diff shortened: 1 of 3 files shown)


jmfee-usgs:
- rebasing should remove the already merged commits.
- it looks like masked streams are being added to merged twice, I suggest something simpler that sorts them at the start:

1) create two streams to hold separate traces:
    merged = Stream()
    masked = Stream()

2) mask each incoming stream, then check each trace:
    add traces with data to merged, 
    add masked traces to masked

3) split and remerge merged

4) add masked to merged (do not trim)

5) unmask everything, and return

geomagio/TimeseriesUtility.py
```diff
@@ -195,17 +195,36 @@ def merge_streams(*streams):
...
     for stream in streams:
         merged += mask_stream(stream)
+
+    # if trace is completely masked separate out to be added back in later
+    for trace in merged:
+        if trace.data.mask.all():
+            if not masked_trace:
+                masked_trace = trace
+            else:
+                masked_trace += trace
```
jmfee-usgs: This is adding two traces, which I believe merges their data even if for separate channels.  I think `masked_trace` should also be a Stream, similar to `merged`.  Maybe add a unit test with 2 empty traces and make sure they are preserved.

test/TimeseriesUtility_test.py
```diff
@@ -113,3 +113,35 @@ def test_get_merged_gaps():
...
+    npts2 = len(trace4.data)
+    timeseries1 = Stream(traces=[trace1, trace2, trace3])
+    timeseries2 = Stream(traces=[trace4, trace5, trace6])
+    for trace in timeseries1:
+        trace.stats.starttime = UTCDateTime('2018-01-01T00:00:00Z')
+        trace.stats.npts = npts1
+    for trace in timeseries2:
+        trace.stats.starttime = UTCDateTime('2018-01-01T00:02:00Z')
+        trace.stats.npts = npts2
+    Merged_streams1 = TimeseriesUtility.merge_streams(timeseries1)
```
jmfee-usgs: do not start method variables with upper case

test/TimeseriesUtility_test.py
```diff
@@ -113,3 +113,35 @@ def test_get_merged_gaps():
...
+    # Make sure the empty 'F' was not removed from stream
+    assert_equals(len(timeseries1), len(Merged_streams1))
```
jmfee-usgs: Select the `F` channel from the merged stream and verify it exists.

test/TimeseriesUtility_test.py
```diff
@@ -113,3 +113,35 @@ def test_get_merged_gaps():
...
+    # Merge multiple streams with overlapping timestamps
+    timeseries = timeseries1 + timeseries2
+    Merged_streams = TimeseriesUtility.merge_streams(timeseries)
+    assert_equals(len(Merged_streams), len(timeseries1))
+    assert_equals(len(Merged_streams[0].data), 6)
+    assert_equals(len(Merged_streams[2]), 6)
```
jmfee-usgs: What are the expected stream values when there are overlaps in the data?  When you change `masked_trace` to a stream, are the overlapping traces still merged, or do they need to be readded before calling `Stream.merge`?

arigdon-usgs:
The masked traces are only being added to merge once (you might be looking at the mask_stream() function at the beginning). The masked_trace is being created at the beginning and the when an entire data set is masked it stores that trace. The split function removes that empty trace from the merged stream while splitting the other traces. The non empty traces are then merged before the masked_trace is added back in to the merged stream.

The trim was to ensure that the masked trace has the same length to the traces that were split() and then merged() while running through the run as update. Without the trim it is appending the large empty set to the smaller sets within the run as update. This does not seem to be breaking anything but I was just trying to make sure the traces within a stream had the same starttime, endtime, and npts. 
Below is a print output from the merged stream just before it is returned by the merge_streams(), this is why I was trimming the trace so that it would have the same parameters as the other traces.

 merged stream after the masked_trace is back and after unmasking 

4 Trace(s) in Stream:
NT.BRW.R0.X | 2018-07-22T04:00:00.000000Z - 2018-07-22T14:00:00.000000Z | 60.0 s, 601 samples
NT.BRW.R0.Y | 2018-07-22T04:00:00.000000Z - 2018-07-22T14:00:00.000000Z | 60.0 s, 601 samples
NT.BRW.R0.Z | 2018-07-22T04:00:00.000000Z - 2018-07-22T14:00:00.000000Z | 60.0 s, 601 samples
NT.BRW.R0.F | 2018-07-22T00:00:00.000000Z - 2018-07-22T23:59:00.000000Z | 60.0 s, 1440 samples
processing 2018-07-22T14:00:01.000000Z 2018-07-22T18:59:59.000000Z ['BRW'] ['X', 'Y', 'Z', 'F']

 merged stream after the masked_trace is back and after unmasking 

4 Trace(s) in Stream:
NT.BRW.R0.X | 2018-07-22T04:00:00.000000Z - 2018-07-22T19:00:00.000000Z | 60.0 s, 901 samples
NT.BRW.R0.Y | 2018-07-22T04:00:00.000000Z - 2018-07-22T19:00:00.000000Z | 60.0 s, 901 samples
NT.BRW.R0.Z | 2018-07-22T04:00:00.000000Z - 2018-07-22T19:00:00.000000Z | 60.0 s, 901 samples
NT.BRW.R0.F | 2018-07-22T00:00:00.000000Z - 2018-07-22T23:59:00.000000Z | 60.0 s, 1440 samples
processing 2018-07-22T19:00:00.000000Z 2018-07-22T23:59:59.000000Z ['BRW'] ['X', 'Y', 'Z', 'F']

 merged stream after the masked_trace is back and after unmasking 

4 Trace(s) in Stream:
NT.BRW.R0.X | 2018-07-22T04:00:00.000000Z - 2018-07-22T23:59:00.000000Z | 60.0 s, 1200 samples
NT.BRW.R0.Y | 2018-07-22T04:00:00.000000Z - 2018-07-22T23:59:00.000000Z | 60.0 s, 1200 samples
NT.BRW.R0.Z | 2018-07-22T04:00:00.000000Z - 2018-07-22T23:59:00.000000Z | 60.0 s, 1200 samples
NT.BRW.R0.F | 2018-07-22T00:00:00.000000Z - 2018-07-22T23:59:00.000000Z | 60.0 s, 1440 samples
processing 2018-07-23T00:00:00.000000Z 2018-07-23T05:00:00.000000Z ['BRW'] ['X', 'Y', 'Z', 'F']

jmfee-usgs:
This looks good, will you add one more test just to be safe:

- One H trace with a gap in the middle
- Another H trace with data that fills the gap, and has an extra sample before/after; but is not as long as the original trace
- Verify the gap was filled as expected.

test/TimeseriesUtility_test.py
```diff
@@ -177,3 +177,14 @@ def test_merge_streams():
     assert_equals(
             numpy.isnan(timeseries3.select(channel='F')[0].data).all(),
             True)
+
+    trace10 = __create_trace('H', [1, 1, numpy.nan, numpy.nan, 1, 1])
+    trace11 = __create_trace('H', [1, 2, 2, 1])
```
jmfee-usgs: Make all of these values in the second trace 2's, so we can see which values are selected after merging

jmfee-usgs: (I suspect it will result in [1, 2, 2, 2, 1, 1])

PR #221, opened 2018-09-20
Trap errors when elevation is not correctly set

### LINKED ISSUE #219: Error when trying to convert a IAGA file into a json file.
I am trying to convert a Hartland IAGA2002 files, that I retrieved from the ~~BGS~~ intermagnet into a json string. Here is an example of one of those files:

```
 Format                 IAGA-2002                                    |
 Source of Data         British Geological Survey                    |
 Station Name           Hartland                                     |
 IAGA CODE              HAD                                          |
 Geodetic Latitude      51.000                                       |
 Geodetic Longitude     355.500                                      |
 Elevation                                                           |
 Reported               HDZF                                         |
 Sensor Orientation                                                  |
 Digital Sampling                                                    |
 Data Interval Type     1-minute                                     |
 Data Type              quasi definitive                             |
 Publication Date       2017-01-13                                   |
 # This data file was created using INTERMAGNET data                 |
 # from the Edinburgh GIN. These data were acquired                  |
 # from an INTERMAGNET quasi-def data file.                          |
 # Final data will be available on the INTERMAGNET DVD               |
 # Go to www.intermagnet.org for details on obtaining this product.  |
 # CONDITIONS OF USE: The conditions of use for data provided        |
 # through INTERMAGNET and acknowledgement templates can be found    |
 # at www.intermagnet.org                                            |
DATE       TIME         DOY     HADH      HADD      HADZ      HADF   |
2017-01-03 00:00:00.000 003     19765.70   -120.67  44368.80  48572.40
2017-01-03 00:01:00.000 003     19765.70   -120.69  44368.80  48572.40
...
```
However, when I try to convert I get an error. Here is my code:
```py
### Loading data in the IAGA 2002 format.

factory = IAGA2002Factory()
data = None
with open('HAD_iaga/had20170101qmin.min', 'r') as f:
    data = factory.parse_string(f.read())
    print(data)

# I don't want to save the output to a file yet, so I save it to a buffer for now.
HAD_buf = io.StringIO("")

output_factory = geomagio.imfjson.IMFJSONWriter()
output_factory.write(
    channels = ('H', 'D', 'Z', 'F'),
    out = HAD_buf,
    timeseries = data)

HAD_str = HAD_buf.getvalue()
HAD_dict = json.loads(HAD_str[2:-1])
HAD_dict
```

And here is the error:

```py
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-33-0ed1e87297ca> in <module>()
     13     channels = ('H', 'D', 'Z', 'F'),
     14     out = HAD_buf,
---> 15     timeseries = data)
     16 
     17 HAD_str = HAD_buf.getvalue()

~/anaconda3/envs/mag_srv_env/lib/python3.6/site-packages/geomagio/imfjson/IMFJSONWriter.py in write(self, out, timeseries, channels, url)
     41         stats = timeseries[0].stats
     42         file_dict['type'] = 'Timeseries'
---> 43         file_dict['metadata'] = self._format_metadata(stats, channels)
     44         file_dict['metadata']['url'] = url
     45         file_dict['times'] = self._format_times(timeseries, channels)

~/anaconda3/envs/mag_srv_env/lib/python3.6/site-packages/geomagio/imfjson/IMFJSONWriter.py in _format_metadata(self, stats, channels)
    124             coords[1] = float(stats.geodetic_latitude)
    125         if 'elevation' in stats:
--> 126             coords[2] = float(stats.elevation)
    127         imo['coordinates'] = coords
    128         intermag['imo'] = imo

ValueError: could not convert string to float: 
```

The error happens when the program tries to convert the elevation of the observatory to a float. This happens because, looking back at the IAGA file, the elevation parameter exists, but is blank.

A possible solution to the problem is to replace the lines (and all similar code):
```py
if 'elevation' in stats:
    coords[2] = float(stats.elevation)
```
with:

```py
try:
    coords[2] = float(stats.elevation)
except KeyError, ValueError, TypeError:
    pass
```
`KeyError` replaces the `if` statement, `ValueError` allows for when the string cannot be converted and `TypeError` allows for when the type is incorrect e.g. `None` type.

jmfee-usgs:
Fixes #219 

Using proposed solution for now, eventually elevation should either not be set or set to `None` when empty.

### DIFF
--- a/geomagio/imfjson/IMFJSONWriter.py
+++ b/geomagio/imfjson/IMFJSONWriter.py
@@ -122,8 +122,11 @@ class IMFJSONWriter(object):
             coords[0] = float(stats.geodetic_longitude)
         if 'geodetic_latitude' in stats:
             coords[1] = float(stats.geodetic_latitude)
-        if 'elevation' in stats:
-            coords[2] = float(stats.elevation)
+        try:
+            if 'elevation' in stats:
+                coords[2] = float(stats.elevation)
+        except (KeyError, ValueError, TypeError):
+            pass
         imo['coordinates'] = coords
         intermag['imo'] = imo
         intermag['reported_orientation'] = ''.join(channels)


erigler-usgs:
Should we apply a similar fix for latitude and longitude?

Arguably, we should do this for any required metadata that isn't a string, but that might be better addressed with an improved metadata model.

jmfee-usgs:
All of those things sound reasonable, this fixes the immediate problem until then.

PR #225, opened 2018-10-01
WIP: Properly handle adjusted/provisional data type

erigler-usgs:
- allow adjusted and provisional strings in --type controller argument
- allow adjusted, provisional, and definitive strings in _get_url() in TimeseriesFactory
- return appropriate abbreviation for adjusted/provisional data type
- translate adjuste/prvisional to A0 location code in EdgeFactory
- only write out DECBAS for variation/reported data types
- translate adjusted/provisional to A0 location code in IMFJSONWriter.py
- in several places I set it up so that 'variation' and 'reported' were synonymous,
  although I am not sure I found every such instance.


erigler-usgs:
@jmfee-usgs, @aclaycomb-usgs,

This works (as Abe saw over my shoulder). I am not certain it addresses every conceivable edge case, but it was nowhere near doing this previously. There is the somewhat larger and unaddressed problem that the IAGA2002 standard defines "provisional" and "variation" data types, while the older IMF standard defines "adjusted" and "reported".

My changes treat these as synonymous, which they are for the most part, but one example of something that is not dealt with is that a user can specify an "adjusted" data type for IAGA2002 output, which is what will get printed in the header, even though this is not technically valid.

PR #229, opened 2018-10-11
Improve --realtime option in Controller.py

erigler-usgs:
The `--realtime` option for the Controller/CLI was previously only a
boolean flag that specified a fixed start/end interval depending on
the interval (minute or second). It still does this, but the user now
has the option to pass an integer argument (e.g., `--realtime 300`)
that allows the user to specify the number of seconds before NOW
to consider "realtime". For a clear explanation of how this works,
see: https://stackoverflow.com/a/34652207.


aclaycomb-usgs:
linting problem only on travis, there may be some need to update.  the 'line break after binary operator' should either show up in all four or none at all

PR #231, opened 2018-10-22
WIP: Filter one minute

aclaycomb-usgs:


geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -0,0 +1,204 @@
+"""Algorithm that converts from one geomagnetic coordinate system to a
+    related geographic coordinate system, by using transformations generated
```
jmfee-usgs: Move the class comments (and update to be correct) to just inside the FilterAlgorithm class docstring

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -0,0 +1,204 @@
...
+        data : numpy.array
+            channel data.
+
+        Returns
+        -------
+        obspy.core.Trace
+            trace containing data and metadata.
+        """
+        stats = Stats(stats)
+        if self.data_type is None:
```
jmfee-usgs: should these be part of the parent class implementation?

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+
+        out = Stream()
+
+        trace_channels = []
+        
+        for trace in stream:
+            trace_channels += trace.stats.channel
+
+        trace_chan_dict = dict(zip(trace_channels, self.outchannels))
```
jmfee-usgs: Does this filter multiple channels at the same time?

Is the stream guaranteed to be in the same order as inchannels (and outchannels)?  Would it be easier to zip inchannels and outchannels for consistency?

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -0,0 +1,204 @@
...
+            stats.pop('processing')
+            stats.npts = filtered.shape[0]
+            trace_out = self.create_trace(
+                stats.channel, stats, filtered)
+
+            out += trace_out
+
+        return out
+    
+    @classmethod
```
jmfee-usgs: why the `@classmethod` annotation?  `cls` appears unused.  You could make it static, or just a method defined outside the class but inside this file/module.

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -0,0 +1,204 @@
...
+            stream containing filtered output
+        """
+        numtaps = len(window)
+
+        # build view into data, with numtaps  chunks separated into 
+        # overlapping 'rows'
+        shape = data.shape[:-1] + (data.shape[-1] - numtaps + 1, 
+                                   numtaps)
+        
+        strides = data.strides + (data.strides[-1],)
```
jmfee-usgs: I think I see what's going on here, and it covers the steps of allowing small gaps, but can you add comments to better explain for future maintainers?

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -0,0 +1,204 @@
...
+            end of input required to generate requested output.
+        """
+
+        half = self.numtaps//2
+        start = start - half*self.interval
+        end = end + half*self.interval
+
+        return (start, end)
+
+@classmethod
```
jmfee-usgs: Align the annotation with the `def` on the following line

geomagio/algorithm/VoltBinAlgorithm.py
```diff
@@ -0,0 +1,179 @@
...
+from __future__ import absolute_import
+
+from .Algorithm import Algorithm
+import json
+import numpy as np
+from obspy.core import Stream, Stats
+import sys
+
+
+class VoltBinAlgorithm(Algorithm):
```
jmfee-usgs: This looks a lot like the Adjusted Algorithm...  is this class ready or can it be removed from this pull request?

geomagio/algorithm/FilterAlgorithm.py
aclaycomb-usgs: it loops over all traces in a stream passed in, applying the filter on each one.  I started with a zip on  inchannels, outchannels, but i didn't do the checks on (H, E, Z, F) -> (MVH, MVE, MVZ, MSF) etc yet.  Will all be easier once the sncl stuff is all standardized.   It appears that the stream will have channels in the order specified in inchannels, but the names of the traces may be the mapped, full 3 letter code.  There maybe problems with all those translations that we don't want to solve here vs the 'right place,' wherever that is...  The new one will I assume map 
(U, V, W, F) to (UFU, UFV, UFW, UFF) for minutes, for example

geomagio/algorithm/FilterAlgorithm.py
aclaycomb-usgs: ok, changed to staticmethod

geomagio/algorithm/FilterAlgorithm.py
aclaycomb-usgs: added more comments to explain the steps, with some mention of python 3.5+ changes coming down the pipe

geomagio/algorithm/FilterAlgorithm.py
aclaycomb-usgs: ok

geomagio/algorithm/VoltBinAlgorithm.py
aclaycomb-usgs: oops.  deleted

erigler-usgs:
Here are my comments after reviewing your responses to jmfee's initial review:

**FilterAlgorithm.py** - This is a good start at the demonstration/validation documentation, but I have some questions and suggestions: 
  1. Why do you need to re-normalize the coefficients obtained using firwin? Is it because you truncate the number of taps at 99? For that matter, why do you NOT re-normalize the coefficients obtained using firwin2?
  2. Are the y-axis labels on the freqz plots correct? Are you reducing power in the stop band by ~1000 dB, or by just a factor of 1000? If the former, wow!
  3. Consider adding freqz plots for the IIR plots, in addition to the group delay plots.
  4. Please include a version of the plot where the filter has been applied, but decimation has not taken place yet. Overplot the original data on the filtered data, and zoom in. I expect this to make immediately obvious what the filter is actually doing by removing (or dampening) what appear to be PC1-PC3 oscillations (~1-45 seconds) in the BOU data. Actually, I see that you sort of did this in the Filter_notes.ipynb file. Maybe copy some of this into FilterAlgorithm.py, since I imagine the *_notes .ipynb file is not intended to be released with the package.
  5. Please add markup and/or comments explaining all cells and plots, and clean up the unused or incomplete material before initial release. Some basic background theory might be nice, at least one or two standard citations should be included, or even just a Wikipedia page if you trust it to be relatively stable. 

**FilterAlgorithm.py** - Just a few suggestions:
  1. Add save_state/load_state methods a la SqDistAlgorithm and AdjustedAlgorithm to read/store filter information. I think this only needs the window, the decimation, and the sample_period, but I'm sure you'll know better than I what the  minimum requirements are to define your FIR filter.
  2. Unless have plans for fancier (i.e., >= "type2") FIR filters in the future, I recommend checking to be sure the window has an odd number of taps, and possibly that it is actually symmetric. The latter is probably less of a concern.
  3. It seems like the code currently implements just a standard Intermagnet 1-minute (from 1-second) Gaussian filter. One way to configure things more flexibly is the JSON state file I just mentioned. Another way is to define command-line arguments that are shortcuts to (as yet unwritten) class/static methods that configure the appropriate filter parameters for pre-determined filters (yes, I saw the `--filter-oneminute` placeholder).

**FilterAlgorithm_test.py** - good start to tests, but I'd like to see a test or two that demonstrate that gappy inputs are handled as you intended. Also, you might consider using NumPy's testing routines (i.e., numpy.testing). What you have is fine, but NumPy's are designed around NumPy arrays, so you wouldn't have to resort to loops to test each output. Your call.

jmfee-usgs:
Merging progress for now, still wip.

PR #232, opened 2018-10-29
WIP: Nix SqDist trim option; modify Controller instead

erigler-usgs:
This is a substantial commit that:

1. reverts a recent change to SqDistAlgorithm that added an option
   to trim trailing NaNs off the input stream; this was in preparation
   for changes needed to fix a long-standing issue with the behavior
   of SqDistAlgorithm when run in the Controller's update mode
   (briefly, it was too easy to trigger a re-initialization of the
   algorithm, which loads 90 days worth of data every time it was
   called, and which ultimately broke the original intended behavior
   of SqDistAlgorithm, which was to run in a time-causal manner, and
   NOT attempt to back-fill arbitrary gaps in real time data).

then...

2. modifies the Controller to support SqDistAlgorithm's requirements
   by a) defaulting to the `next_starttime` stored in each instance
   of SqDistAlgorithm when calling the `get_input_interval` member
   function; and b) trimming any trailing NaNs off of the ends of
   inputs retrieved for the end-gaps in any update interval. Together,
   these changes ensure that the time-causal assumptions inherent
   to SqDistAlgorithm are always satisfied unless the user very
   deliberately removes the `statefile` that holds `next_starttime`
   between calls.

Long-story-short, there was no obvious way to get the desired behavior
from SqDistAlgorithm without modifying the Controller. These changes
do NOT break any existing algorithms, which all simply return the
starttime and endtime passed to `get_input_interval` member function.
Of course if any future algorithms need to return intervals that are
different from the inputted starttime and endtime, they will need to
be tested carefully against this new Controller logic, but I don't
anticipate problems unless they incorporate a fundamentally different
concept of time dependence than SqDistAlgorithm.

Finally...

3. Some simple tests were added to SqDistAlgorithm to make sure that
   it's `get_input_interval` behaves as expected. Some not-so-simple
   tests were added to the Controller that make certain the logic
   required for SqdistAlgorithm holds. These tests require multiple
   test output files, one for each of a sequence of calls to the
   Controller in its update mode. This in turn makes heavy use of
   the IAGA2002Factory, and the input and output URL functionality
   built into geomag-algorithms. In short, this comprised an end-
   to-end test of one of the more complicated algorithms and use-
   cases expected to be encountered by geomag-algorithms. Certainly
   not what is usually considered a 'unit' test, but very useful.

-


erigler-usgs:
I deliberately named both @jmfee-usgs  and @aclaycomb-usgs as reviewers.

@jmfee, Will you look at this to see if the changes are appropriate for the Controller. I can't think of a better way to do this, but you might have some different ideas. Hopefully the new test outputs will make it clear exactly what I need the SqDistAlgorithm to do, but if not, let's discuss.

@aclaycomb-usgs, Will you look at this with an eye toward seeing if this will break your new filter algorithm. I don't think it will, but I know you make use of the `get_input_interval()` to return an interval that differs from the actual gap you are trying to update, so this assessment is not as straight-forward as with other existing algorithms.

Thanks.

erigler-usgs:
OK, @aclaycomb-usgs  sat down with  me to look closely at these changes, and we determined that my the modified Controller will indeed break his new filter algorithm. Since the new filter algorithm is needed for more critical low-level Geomag data products, I'm going back to the drawing board and will try harder to make sure any modifications do not break existing functionality. Thus the change in title for this PR to no include "WIP:".

Please do not merge without consulting with me first. Thanks.

jmfee-usgs:
@erigler-usgs : I'll look at this more carefully, but can you describe why this isn't working with get_input_interval's start and next_starttime?  Can you just change that to not require strict equality?  Or is this an issue with how the gaps are computed across dst/sq/sv?

erigler-usgs:
@jmfee-usgs, I'm not sure what you mean by "just change that to not require strict equality". Unless you are referring to the old version (pre-#232). I changed get_input_interval() to always defer to next_starttime unless it was None, in which case it would return a starttime required to re-initialize. For the most part, this addresses the problem of too frequent re-initialization.

But, I think the second part of your question is getting more to the heart of my problem. My desire is that SQ and SV have no gaps *except* the open-ended gap associated with the real time end of the input stream. Dist, on the other hand, should have gaps that fully reflect the real time input stream, with "real time" being configurable. The biggest hurdle to getting exactly what I want with a state-dependent algorithm seems to be gaps that span (potentially multiple) update interval boundaries.

Anyway, the initial (WIP) PR works perfectly for SqDistAlgorithm, but as I mentioned above, it will break other algorithms that also require inputs that differ from the requested output times. I'm starting to think I need to modify the base Algorithm class to include another function akin to can_produce_data(), but one which can actually massage the input stream to satisfy each Algorithm's specific requirements. I was trying to avoid this, but as I consider it more, I think this might actually require a smaller change than my initial PR, at least to the Controller.

Stay tuned...

erigler-usgs:
OK @jmfee-usgs and @aclaycomb-usgs, 

I updated this PR to migrate the new Controller functionality to an SqDistAlgorithm class "pre-process" function, and created pass-through stubs for the parent Algorithm class. It's not very elegant, but it serves my purpose (all previous end-to-end tests pass), and it does not break existing or anticipated new algorithms. Please let me know what you think. Thanks.

jmfee-usgs:
Replaced by #234

PR #234, opened 2018-11-01
better support for stateful algorithms

jmfee-usgs:
Please do not merge yet, this is the start of an update to SQDist to address reinitialization issues.

- Add "get_starttime" algorithm API method
- limited integration to controller


jmfee-usgs:
@erigler-usgs : this is very preliminary, but I think covers what we discussed (after TODOs and such are implemented).  Does this seem like a good direction?

geomagio/Controller.py
```diff
@@ -158,11 +160,13 @@ def run(self, options, input_timeseries=None):
                 algorithm.get_input_channels()
         output_channels = options.outchannels or \
                 algorithm.get_output_channels()
+        starttime = algorithm.get_starttime() or options.starttime
+        endtime = options.endtime
```
erigler-usgs: If this is specific to moving forward only (i.e., time-dependent stateful algorithms), maybe a little more descriptive name, like "get_next_starttime". Not only is this applicable to any forward-propagating algorithm like SqDistAlgorithm, but it is not inconsistent with the way you've defined "gaps" elsewhere (to be clear, I'm not suggesting you could use that last item to your advantage here, just pointing it out).

geomagio/algorithm/SqDistAlgorithm.py
```diff
@@ -194,12 +199,20 @@ def process_one(self, trace):
             if trace.stats.station != self.last_observatory \
                     or trace.stats.channel != self.last_channel \
                     or trace.stats.starttime != self.next_starttime:
+                # TODO: raise exception to prevent state from being cleared?
```
erigler-usgs: Yes. Additionally, we need to add a test to confirm that the trace.stats['delta'] values are compatible. This requires adding delta to the state, which might break existing software, but all would be fixed within a single re-initialization.

geomagio/algorithm/SqDistAlgorithm.py
```diff
@@ -194,12 +199,20 @@ def process_one(self, trace):
...
                 # state not correct, clear to be safe
                 self.yhat0 = None
                 self.s0 = None
                 self.l0 = None
                 self.b0 = None
                 self.sigma0 = None
+
+        # TODO: prepare data for processing
+        # - controller/get_starttime should ensure state is honored
+        # - based on a TBD argument (default 30 minutes)
```
erigler-usgs: Can't this just use the existing `--realtime` option that I modified recently to accept an integer number of seconds as an argument. This just sets options.starttime, which as far as I can tell, is never modified except in the Controller's run_as_update().

geomagio/algorithm/SqDistAlgorithm.py
```diff
@@ -194,12 +199,20 @@ def process_one(self, trace):
...
+        #   project SQ/SV up to (default 30 minutes) behind realtime
+        #   by padding end with nan values for additive call.
+
         # trim trailing NaNs if self.trim is set
         if self.trim:
             trace.stats['npts'] = np.argwhere(~np.isnan(trace.data))[-1][0] + 1
```
erigler-usgs: The reason I used the more cryptic reverse generator in my previous PR (instead of argwhere) was that I expect there will be far more valid points at earlier times than NaNs at later times. Maybe only important for long traces, but I see that as being a common use-case. Anyway, yours is certainly more readable.

geomagio/algorithm/SqDistAlgorithm.py
```diff
@@ -229,6 +242,8 @@ def process_one(self, trace):
         self.sigma0 = sigma0
         self.last_observatory = trace.stats.station
         self.last_channel = trace.stats.channel
+        # TODO: double check next_starttime.  Does this work as expected when
+        # projecting into the future, as long as the inputs are set correctly?
```
erigler-usgs: Isn't this already handled at the beginning of process_one() (with on or two minor tweaks I already mentioned above), when the state is tested? Or am I not understanding your question here?

geomagio/Controller.py
jmfee-usgs: I've made this suggested change.

geomagio/algorithm/SqDistAlgorithm.py
jmfee-usgs: I think so, just leaving a note to make sure this is tested

geomagio/algorithm/SqDistAlgorithm.py
jmfee-usgs: Yes, I think that's reasonable.

When not running in --realtime mode, should it default to how --sqdist-trim works?  Should we set a default and/or update how --sqdist-trim works?

erigler-usgs: I forgot you were starting from a version where --sqdist-trim still existed. I'll have to think about this more. Stay tuned...

erigler-usgs: So, short answer as per our Google-chat...if you weren't already planning to use `--sqdist-trim`, feel free to get rid of it, and know that you have the `fc` option in SqDistAlgorithm.additive() that you might also be able to use to your advantage.

jmfee-usgs:
@aclaycomb-usgs : please take a look at these changes.

The Controller_test includes tests from #232 , updated to not use "--update".  I also modified "test 5" to check the controller behavior for projecting up to the "realtime gap".

jmfee-usgs:
(I think `TimeseriesUtility` needs a few more tests before merging)

Issue #239, opened 2019-02-06
SqDistAlgorithm.process_one() problem with empty trace


erigler-usgs:
When SqDistAlgorithm.additive() (the heart of SqDistAlgorithm) receives a zero-length array-like input, it returns a zero-length array-like object, and correctly leaves all initial values unchanged. However, the wrapper SqDistAlgorithm.process_one() incorrectly updates the `next_starttime` state parameter.

https://github.com/usgs/geomag-algorithms/blob/811adb911741fea7b8bb70a9b992804b8682dff6/geomagio/algorithm/SqDistAlgorithm.py#L261

This is due to weirdness in the obspy.core.Trace class, wherein you can set the number of points to zero, but then the `Trace.stats.starttime` and `Trace.stats.endtime` are not well-defined.

https://github.com/usgs/geomag-algorithms/blob/811adb911741fea7b8bb70a9b992804b8682dff6/geomagio/TimeseriesUtility.py#L384-L388

The solution I propose is to check the length of the array returned from SqDistAlgorithm.additive(), and multiply this by `delta` before adding to the existing `next_starttime` state parameter. This is a bit of a kludge, but I would argue that the obspy.Trace data model is just not sufficient to deal with this situation, and we should not waste effort to try to get it to do the right thing.

@jmfee-usgs, I'd appreciate your take on this before I make this change and push out a new release. FYI, the practical problem this causes is that `next_starttime` gets incremented even when no processing is done because the input time series got trimmed to zero-length to accommodate the `realtime_gap` calculated in the Controller.

jmfee-usgs:
Would it be easier to skip process_one when `trace.npts == 0`?

Alternatively, the calculation could use npts instead of adding to endtime:
```
self.next_starttime = trace.stats.starttime + (trace.stats.delta * trace.stats.npts)
```

erigler-usgs:
A reason **not** to skip process_one() is that the additive function has the ability to forecast into the future. Actually, there are two ways it can do this: One time-shifts the inputs vs. outputs using the `hstep` parameter; and one actually simulates `fc` discrete steps beyond the last input by propagating the state forward in time. 

We don't use either of these currently, and probably need to document this functionality better before making it available through the Controller, but I'd rather not make such future changes more difficult than they need to be.

So, this requires that we do **not** rely on trace.stats.npts. What I think we need is:

```python
self.next_starttime = self.next_starttime + trace.stats.delta * (len(yhat) + self.hstep)
```

[...]

PR #240, opened 2019-02-12
Leave SqDistAlgorithm's state alone if no inputs

erigler-usgs:
Fixes issue #239, whereby the next_starttime state variable was
incremented even if no input was passed to SqDistAlgorithm.process_one(),
and therefore no processing was performed by SqDistAlgorithm.additive().
Also, modified SqDistAlgorithm.additive() so that the state variables
returned were consistent with the last *input* sample, and NOT with the
last output sample (they could differ if fc input parameter were ever
used).


erigler-usgs:
I re-assigned this to @aclaycomb-usgs, but @jmfee-usgs looked at this with me, and while there were many suggestions for future improvements, seemed OK to see this PR go forward.

PR #246, opened 2019-05-23
Pcdcp one second factory ingest

aclaycomb-usgs:


geomagio/pcdcp/PCDCPFactory.py
```diff
@@ -54,20 +54,15 @@ def parse_string(self, data, **kwargs):
         parser = PCDCPParser()
         parser.parse(data)
 
-        yr = int(parser.header['year'])
-        yrday = int(parser.header['yearday'])
+        sample_periods = {4: 60.0, 5: 1.0}
```
erigler-usgs: This is a little obscure. Insert a brief comment explaining why the dictionary keys are '4' and '5'.

erigler-usgs:
This looks good once you add the requested comment on the logic used to differentiate minute from second PCDCP files. I just want to point out something related to the perceived need to explicitly name the channels in the test routines for consistency between OS's.

The PCDCPParser._post_process() method creates a dictionary with keys named after the channels created in PCDCPParser._set_channels(). Dictionary keys are **not** guaranteed to be stored in the order they were created in Python. HOWEVER, ObsPy Streams, which are what are used in the test routines, are not based on dictionaries, but are rather glorified **lists* of ObsPy Traces.

So, if one wished to avoid such problems in the future, they would not store data read in by the Parser as a dictionary indexed by keys, but rather as a list of lists. After all, the channel names are already stored in the Parser object, and in the desired order.

I'm not proposing that you do this, I just wanted to be sure we understood the source of the discrepancies that arose when using Linux to test code developed on MacOS.

geomagio/pcdcp/PCDCPFactory.py
```diff
@@ -53,20 +54,17 @@ def parse_string(self, data, **kwargs):
         parser = PCDCPParser()
         parser.parse(data)
 
-        yr = int(parser.header['year'])
-        yrday = int(parser.header['yearday'])
+        # minutes files times are 4 characters long (1440)
+        # seconds files times are 5 characters long (86400)
```
aclaycomb-usgs: added these lines to address dictionary for selecting seconds/minutes

geomagio/pcdcp/PCDCPFactory.py
aclaycomb-usgs: https://github.com/usgs/geomag-algorithms/pull/246/files#r288666760

geomagio/pcdcp/PCDCPFactory.py
erigler-usgs: Perfect.

PR #247, opened 2019-06-07
Sncl translation

jmfee-usgs:


geomagio/miniseed/sncl.py
```diff
@@ -0,0 +1,149 @@
...
+    if location_suffix in SEED_LOCATION_SUFFIX:
+        element_suffix += '_' + SEED_LOCATION_SUFFIX[location_suffix]
+    return element_prefix + element_suffix
+
+
+def get_sncl_channel(network, observatory, element, data_interval, data_type):
+    channel_prefix = INTERVAL_CHANNEL_PREFIX[data_interval]
+    element_parts = element.split('_')
+    channel_suffix = element_parts[0]
+    channel_middle = 'F'
```
erigler-usgs: I haven't processed the entire PR yet, so maybe this is a non-issue, but I anticipate using geomag-algorithms to handle more than just magnetic data ('F' instrument code). For example, 'Q' for electric fields. Also, what about housekeeping codes like battery voltage and temperature?

erigler-usgs:
@jmfee-usgs: I started to review this, but am now thinking this is more of a WIP. Should I hold off for now?

jmfee-usgs:
@erigler-usgs I agree additional codes will be needed in the future, but this implements what we can expect from the updated ObsRIO configuration.  We can add additional codes later when we're ready...

jmfee-usgs:
Closing for now, miniseed support was implemented elsewhere.

Issue #121, opened 2016-07-11
Make certain .tmp directory exists before writing file.


hasimpson-usgs:
We write files out for dcpmessages to read in.  If the directory doesn't exist,  it fails ungracefully.

  File "/home/geomag_user/geomagio/geomag-algorithms/geomagio/imfv283/GOESIMFV283Factory.py", line 185, in _fill_criteria_file
    with open(criteria_file, 'wb') as fh:
IOError: [Errno 2] No such file or directory: './.tmp/TUC.sc'

PR #252, opened 2019-07-02
Fixes issue 121

pcain-usgs:
Verifies whether or not the filename created exists. If not, the function returns with no output. If the filename exists, an outpupt file is written.

### DIFF
--- a/geomagio/imfv283/GOESIMFV283Factory.py
+++ b/geomagio/imfv283/GOESIMFV283Factory.py
@@ -5,7 +5,7 @@ from .IMFV283Factory import IMFV283Factory
 import subprocess
 import sys
 from obspy.core import Stream
-
+import os
 
 class GOESIMFV283Factory(IMFV283Factory):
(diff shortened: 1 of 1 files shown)


geomagio/imfv283/GOESIMFV283Factory.py
```diff
@@ -181,7 +180,7 @@ def _fill_criteria_file(self, starttime, endtime, observatory):
         buf.append('RETRANSMITTED: N\n')
         buf.append('ASCENDING_TIME: false\n')
         buf.append('RT_SETTLE_DELAY: true\n')
-
+        if os.path.exists(criteria_file) is not True: return
```
jmfee-usgs: - always indent statements that are part of an if
- `X is not True` should be written as `not X`

- it seems like if the file doesn't exist, this prevents it from being written?
- #121 suggests the directory needs to be created, so you might want something like:

```python
criteria_dir = os.path.dirname(criteria_file)
if not os.path.exists(criteria_dir):
  os.makedirs(criteria_dir)
# create file as usual
```

PR #262, opened 2020-01-02
Flask webservice

jmfee-usgs:
Initial implementation for flask webservice with oauth login.

Proposed metadata model for flagging, etc.


jmfee-usgs:
@erigler-usgs : for discussion next week.

I had to make a small change to SQDist based on an expected numpy parameter type for tests to pass.

Take a look at the webservice/Metadata.py.  I think this covers a general purpose metadata schema.  We will still need to nail down any category-specific details for adjusted matrices, etc, that would be stored as json.  These details wouldn't be directly searchable, and you would query based on a time range and category; and search/filter results on the client side...
https://github.com/usgs/geomag-algorithms/pull/262/files#diff-8e6d3cb1b2d2ae9f6443019bad619fe1

PR #265, opened 2020-01-28
FilterAlgorithm

pcain-usgs:
Algorithm to filter data from 10Hz to specified output sample period. Also included is an updated unit test with the required iaga files and a set of coefficients to test custom filter functionality.


geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -73,27 +95,77 @@ def process(self, stream):
...
+            self.numtaps = len(self.window)
+            for trace in stream:
+                data = trace.data
+                step = int(output_sample_period /
+                    input_sample_period)
+                filtered = self.firfilter(data, self.window, step)
+                stats = Stats(trace.stats)
+                self.numtaps = len(self.window)
+                stats.starttime = trace.stats.starttime + \
+                self.numtaps * self.input_sample_period // \
```
jmfee-usgs: line continuations should be indented by 8 spaces, to make it clearer.

Operators should be placed at the start of the continuation line.  Make sure to group related operations for clarity, e.g. (numtaps * input_sample_period // 2)

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -73,27 +95,77 @@ def process(self, stream):
...
+                2 + self.input_sample_period
+                stats.delta = stats.delta * step
+                stats.npts = filtered.shape[0]
+                trace_out = self.create_trace(
+                    stats.channel, stats, filtered)
 
+                out += trace_out
+            return out
+        # initialize filter to execute cascading filtering steps
+        self.window = [sps.firwin(123, 0.45 / 5.0, window='blackman'),
```
jmfee-usgs: I like the cascading filter, but I think this needs to be reorganized to be clearer.

Group the related parameters for each step, e.g:
```
step = {
  input_sample_period: 0.1,
  output_sample_period: 1.0,
  window: sps.firwin...,
  decimation: 10,
  ...
```

These parameters for a given step are the same as how the custom steps are applied.  The predefined steps should be defined as variables outside the class.

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -198,4 +327,27 @@ def configure(self, arguments):
...
         """
-        pass
+        Algorithm.configure(self, arguments)
+
+        self.filtertype = arguments.filter_type
+        self.volt_conv = arguments.volt_conversion
+        self.bin_conv = arguments.bin_conversion
+        self.window = arguments.filter_coefficients
+
+        if arguments.input_interval in ['tenhertz']:
```
jmfee-usgs: These should use the utility method get_delta_from_interval:

https://github.com/usgs/geomag-algorithms/blob/99e47f514600c0ce177f700d14a8985865c1bc5b/geomagio/TimeseriesUtility.py#L56

test/algorithm_test/FilterAlgorithm_test.py
```diff
@@ -8,36 +10,142 @@ def test_process():
...
-    min_iaga2002_string = min_iaga2002_file.read()
-    min_iaga2002_file.close()
-    sec_iaga2002_file = open('etc/filter/BOU20180901vsec.sec')
+    # initialize 10Hz to 1sec filter
+    f = FilterAlgorithm(filtertype='default',
+                        input_sample_period=0.1,
+                        output_sample_period=1)
+    starttime = UTCDateTime('2020-01-06T00:00:00Z')
+    endtime = UTCDateTime('2020-01-07T00:00:00Z')
+    starttime, endtime = f.get_input_interval(starttime, endtime)
```
jmfee-usgs: @pcain-usgs , @erigler-usgs this host name needs to be removed.  You can store miniseed or other files in this directory if that helps for testing purposes.  This needs to be done in a way that rewrites history.

Also, it looks like the factory setup code could be shared.

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -73,27 +95,77 @@ def process(self, stream):
         out : obspy.core.Stream
             stream containing 1 trace per original trace.
         """
-
+        # if input stream is 10 Hz, convert data to nT
+        if self.input_sample_period == 0.1:
+            stream = self.convert_miniseed(stream)
```
jmfee-usgs: Rather than implying the input_sample_period should be converted from miniseed, change this to be more explicit.

There are the --volt-conversion NT/V and --bin-conversion NT/B arguments.

Add a --convert-voltbin CHANNELS argument, e.g. `--convert-voltbin U V W`, that specifies which channels to convert.  When this argument is set, pass that list of channels to be converted along with the conversion factors.

Move the convert_miniseed function outside this class and make it take a stream and the other arguments; also rename to convert_voltbin.  Inside that function, you can use the input channel name to look for Volt/Bin channels to be converted and skip the sorting.

geomagio/edge/MiniSeedFactory.py
```diff
@@ -127,15 +132,27 @@ def get_timeseries(self, starttime, endtime, observatory=None,
...
         finally:
             output = temp_stdout.getvalue()
             if output:
                 sys.stderr.write(str(output))
             temp_stdout.close()
             sys.stdout = original_stdout
-        self._post_process(timeseries, starttime, endtime, channels)
 
+        if self.convert_channels is not None:
+            out = obspy.core.Stream()
```
jmfee-usgs: This still depends on careful input argument order.  It would be more robust to explicitly select the necessary channels:

```
for channel in self.convert_channels:
    in = timeseries.select(channel=channel+'_Bin') \
            + timeseries.select(channel=channel+'_Volt')
    out += self.convert_voltbin(channel, in, self.bin_conv, self.volt_conv)
```

And:

```
def convert_voltbin(self, channel, stream, nt_per_bin=500, nt_per_volt=100):
    bins = stream.select(channel=channel+'_Bin')
    volts = stream.select(channel=channel+'_Volt')
    ...
```

geomagio/edge/MiniSeedFactory.py
```diff
@@ -448,6 +465,39 @@ def _get_timeseries(self, starttime, endtime, observatory,
...
+        # selects volts from input Trace
+        volts = stream.select(channel="*V*")
+        # selects bins from input Trace
+        bins = stream.select(channel="*B*")
+        # copy stats from original Trace
+        stats = obspy.core.Stats(volts[0].stats)
+        # set channel parameter to U, V, or W
+        stats.channel = volts[0].stats.channel[0]
+        # conversion from bins/volts to nT
+        data = int(self.volt_conv) * \
```
jmfee-usgs: The conversion factors should be floats.

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -1,38 +1,88 @@
...
+            decimation=None, steparr=None, volt_conv=None, bin_conv=None,
+            inchannels=None, outchannels=None):
+
+        Algorithm.__init__(self, inchannels=None, outchannels=None)
+        self.window = window
+        self.steparr = steparr
+        self.coeff_filename = coeff_filename
+        self.filtertype = filtertype
+        self.numtaps = numtaps
+        self.volt_conv = volt_conv
```
jmfee-usgs: Rather than setting this and setting a default below, you can either set the default in the parameters list or something like this:

```
self.volt_conv = volt_conv or 100.0
```

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -1,38 +1,88 @@
...
+        self.bin_conv = bin_conv
+        self.input_sample_period = input_sample_period
+        self.output_sample_period = output_sample_period
+
+        # Initialize filter to execute filtering step with custom coefficients
```
jmfee-usgs: I recommend a separate method:

```
get_filter_steps(self):
    if self.filtertype == 'custom':
        step = {
            'input_sample_period': self.input_sample_period,
            'output_sample_period': self.output_sample_period,
            'window': self.window,
            ...
        }
        return [step]
    steps = []
    for step in STEPS:
        if self.input_sample_period <= step.input_sample_period
                and self.output_sample_period >= step.output_sample_period:
            steps.append(step)
    return steps
```

Then your other methods can just loop over steps and access current step properties.  The STEPS variable would be a constant outside the class and define the 10Hz->1 second, etc, steps that are currently in parallel arrays.

geomagio/edge/MiniSeedFactory.py
jmfee-usgs: I made the conversion factors separate parameters because we might store multiple conversion factors and call this method for multiple channels using different values...

geomagio/edge/MiniSeedFactory.py
```diff
@@ -448,6 +465,39 @@ def _get_timeseries(self, starttime, endtime, observatory,
...
+            volts[0].data + int(self.bin_conv) * bins[0].data
+        # create empty trace with adapted stats
+        out = TimeseriesUtility.create_empty_trace(stats.starttime,
+                stats.endtime, stats.station, stats.channel,
+                stats.data_type, stats.data_interval,
+                stats.network, stats.station, stats.location)
+        # set data for empty trace as nT converted data
+        out.data = data
```
jmfee-usgs: Double check whether you also need to set `stats.npts`

geomagio/algorithm/FilterAlgorithm.py
pcain-usgs: I just pushed an alternative fix. The steps were put into a separate json file, where load state creates an environment(dictionary) for the filter to operate with(sets window, decimation, input/output sample periods, etc.). This greatly simplifies the process method while allowing it to use one loop for either a default or custom filtering operation.

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -73,27 +136,34 @@ def process(self, stream):
...
-            out += trace_out
-
+        # intitialize dictionary object for filter
+        steps = self.get_filter_steps()
+        out_stream = Stream()
+        out_stream = stream.copy()
+        # while dictionary is not empty
+        for step in steps:
+            # gather variables from single step
+            window = np.array(step['window'])
```
jmfee-usgs: Move the inner loop to a method `process_step(self, step, stream)` that returns the processed stream.  Then you can say `stream = process_step(self, step, stream.copy())`, and return stream at the end...

A few more requests:
- Read all the variables from the step first, then the calculated variables (alphabetical within each section).
- Should `window / sum(window)` be calculated once before the trace loop?
- Not sure the `data` variable is needed
- is filtered.shape[0] the same as len(filtered)?
- I think you only need to set `delta`, and it automatically generates `sampling_rate`.

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -163,19 +229,19 @@ def get_input_interval(self, start, end, observatory=None, channels=None):
...
-
-        half = self.numtaps // 2
-        start = start - half * self.sample_period
-        end = end + half * self.sample_period
-
+        steps = self.get_filter_steps()
+        # calculate start/end from step array
+        for step in steps:
+            half = len(step["window"]) // 2
+            start = start - half * step["input_sample_period"]
```
jmfee-usgs: calculate `half_step` once, then add/subtract

geomagio/edge/MiniSeedFactory.py
```diff
@@ -448,6 +461,41 @@ def _get_timeseries(self, starttime, endtime, observatory,
...
+        channel: string
+            channel string(U ,V ,W)
+        Returns
+        -------
+        out : obspy.core.Trace
+            Trace containing 1 trace per 2 original traces.
+        """
+        out = obspy.core.Trace()
+        # selects volts from input Trace
+        volts = stream.select(channel=channel + "_Volt")
```
jmfee-usgs: Since this only processes the first trace, you could add that here so it doesn't repeat below
```
volts = stream.select(channel=channel + "_Volt")[0]
```

geomagio/edge/MiniSeedFactory.py
```diff
@@ -448,6 +461,41 @@ def _get_timeseries(self, starttime, endtime, observatory,
...
+        # selects bins from input Trace
+        bins = stream.select(channel=channel + "_Bin")
+        # copy stats from original Trace
+        stats = obspy.core.Stats(volts[0].stats)
+        # set channel parameter to U, V, or W
+        stats.channel = channel[0]
```
jmfee-usgs: This should be `channel` now, instead of `channel[0]`

geomagio/edge/MiniSeedFactory.py
```diff
@@ -448,6 +461,41 @@ def _get_timeseries(self, starttime, endtime, observatory,
...
+        # conversion from bins/volts to nT
+        data = self.volt_conv * \
```
jmfee-usgs: if the line continuation is needed, break between separate groups, e.g.

```
data = self.volt_conv * volts.data \
        + self.bin_conv * bins.data
```

test/algorithm_test/FilterAlgorithm_test.py
```diff
@@ -8,36 +9,156 @@ def test_process():
...
-                         outchannels=('H', 'E', 'Z', 'F'))
-
-    filt_bou = a.process(sec)
-
-    # unpack channels from loaded minutes data file
-    u = mint.select(channel='H')[0]
-    v = mint.select(channel='E')[0]
-    w = mint.select(channel='Z')[0]
-    f = mint.select(channel='F')[0]
+    # initialize 10Hz to 1sec filter
```
jmfee-usgs: Separate these into three tests, one for each filter algorithm configuration

test/algorithm_test/FilterAlgorithm_test.py
```diff
@@ -8,36 +9,156 @@ def test_process():
     Check one minute filter data processing versus files generated from
```
jmfee-usgs: Make the first line of the docstring the path to the function within the test directory, since this is output by nose and helps find a test that has errors:

```
"""algorithm_test.FilterAlgorithm_test.test_process()

Any other comments (if needed).
"""

PR #268, opened 2020-02-13
Various changes

jmfee-usgs:
- Fix stdout/stderr buffering to capture obspy stdout.
- Switch back to obspy based earthworm client (had custom version while pr was pending)
- Fix weird data type error observed in EdgeFactory with data gaps
- Reorganize controller argument parsing, clean up usage


geomagio/Controller.py
```diff
@@ -537,9 +471,13 @@ def main(args):
     if args.output_observatory is None:
         args.output_observatory = args.observatory
     elif args.observatory_foreach:
-        raise Exception("Cannot specify" +
+        raise Exception("Cannot combine" +
              " --output-observatory and --observatory-foreach")
 
+    if args.output_stdout and args.update:
+        raise Exception("Cannot combine" +
```
erigler-usgs: Am I correct in thinking that the miniseed factory is, at least currently, not capable of writing? Should we add something here about that too?

jmfee-usgs: I added most of the support for that previously, and forgot to configure it within the factory. 
 We'll need to confirm the actual ports being used, I think that varies much more than the read port, and confirm the protocol works as expected (write miniseed formatted data to a socket).

erigler-usgs: To be clear, I wasn't suggesting we activate the miniseed write capability (which seems to still require a bit more development and testing). I was suggesting that we generate a message that writing to miniseed was not currently possible, similar to what was done when stdout was specified as the output.

But if you think this is functionality that will soon be available, never mind. I will merge.

PR #269, opened 2020-02-13
WIP: Adds support for FGE volts channels.

pcain-usgs:
Edits metadata with FGE and scaling parameters. Edits miniseed factory to comprehend FGE vs NAROD sensors. Simplifies coefficient calculation in Filtering Algorithm. Passes all tests with the exception of the flask test. WIP because there is an obspy problem when attempting to retrieve data in an API setting.


pcain-usgs:
> WIP because there is an obspy problem when attempting to retrieve data in an API setting.

Note that this issue only occurs when gathering data from BDT

geomagio/Metadata.py
```diff
@@ -0,0 +1,90 @@
...
+                "V": [{"channel": "V_Volt", "offset": 0, "scale": 312.3}],
+                "W": [{"channel": "W_Volt", "offset": 0, "scale": 312.0}],
+            },
+            "electronics": {
+                "serial": "E0542",
+                # these scale values are used to convert voltage
+                "x-scale": 313.2,  # V/nT
+                "y-scale": 312.3,  # V/nT
+                "z-scale": 312.0,  # V/nT
+                "temperature-scale": 0.01,  # V/K
```
erigler-usgs: These units (in the comment) should be "nT/V" and "K/V".

erigler-usgs: Will fix with new pull request.

PR #271, opened 2020-02-18
Flask webservice

trivers-usgs:


geomagio/webservice/data.py
```diff
@@ -0,0 +1,195 @@
...
+        return render_template('usage.html')
+
+    query = parse_query(query_params)
+
+    timeseries = get_timeseries(query)
+
+    return format_timeseries(timeseries, query)
+
+
+def parse_query(query):
```
jmfee-usgs: Copy the WebServiceQuery object ( from https://github.com/usgs/geomag-algorithms/blob/master/geomagio/WebService.py#L340 ) into this module.

Update this method to be more like the existing parse (order, etc), just using flask args instead of parse_qs ( https://github.com/usgs/geomag-algorithms/blob/master/geomagio/WebService.py#L255 ), and make it a class method that returns a parsed WebServiceQuery object.

Defaults should be constants at the top of the file, instead of inline values.

geomagio/webservice/data.py
```diff
@@ -0,0 +1,195 @@
...
+
+    # Get data type
+    params['Type'] = query.get('type')
+
+    validate_parameters(params)
+
+    return params
+
+
+def validate_parameters(params):
```
jmfee-usgs: Use the existing WebServiceQuery validate method (as an instance method) and move the valid lists to constants at the top of the file.

geomagio/webservice/data.py
```diff
@@ -0,0 +1,195 @@
...
+    ----------
+    query: dict
+        dictionary of parsed query parameters
+
+    Returns
+    -------
+    obspy.core.Stream
+        timeseries object with requested data
+    """
+    factory = EdgeFactory()
```
jmfee-usgs: The factory should allow configuration using environment variables (INPUT_TYPE, INPUT_HOST, INPUT_PORT, etc).  This is essentially the same as `fetch` ( https://github.com/usgs/geomag-algorithms/blob/master/geomagio/WebService.py#L136 ).

Add a `get_interval_from_delta` method to TimeseriesUtility to handle the sampling_period to interval logic, the reverse of `get_delta_from_interval` ( https://github.com/usgs/geomag-algorithms/blob/master/geomagio/TimeseriesUtility.py#L56 )

geomagio/webservice/data.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+    if query['Format'] == 'json':
+        json_output = IMFJSONWriter.format(timeseries, query['Channels'])
+
+        return json_output
+
+    else:
+        iaga_output = IAGA2002Writer.format(timeseries, query['Channels'])
+
+        iaga_output = Response(iaga_output, mimetype="text / plain")
```
jmfee-usgs: This mimetype should be `text/plain` (no spaces).  Set the json mime type to `application/json`

geomagio/webservice/data.py
```diff
@@ -0,0 +1,195 @@
...
+def get_data():
+    query_params = request.args
+
+    url = request.url
+
+    if not query_params:
+
+        return render_template('usage.html')
+
+    query = parse_query(query_params)
```
jmfee-usgs: wrap these statements in a try block, and add error formatting if an exception occurs.  Generally WebServiceException errors should return a `400 Bad Request` http response with details, other errors should return a `500 Internal Server Error` http response and be printed to server logs.  See the existing error handling for details ( https://github.com/usgs/geomag-algorithms/blob/master/geomagio/WebService.py#L190 )

geomagio/webservice/templates/usage.html
```diff
@@ -0,0 +1,683 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML+RDFa 1.0//EN"
```
jmfee-usgs: Move this into `templates/data/usage.html`.
The template should extend `_template.html` (similar to hello or index), and only include usage details in the content block.

The template can be updated to use the wret app layout later.

geomagio/webservice/data.py
trivers-usgs: Which class are you thinking the parse method should be a part of?  I played with having it in the `WebServiceQuery`, but that was a bit of an awkward implementation.

jmfee-usgs: It _could_ be a `@staticmethod` on WebServiceQuery, but it's okay where it is.  If we leave it where it is, I recommend also removing the _validate_parameters method from the class for consistency (move it to `validate_query(query)` after parse_query at the end).

The parse method still uses different logic from the original implementation.  For this to be correct, it needs to at least check and set a starttime default before handling endtime:
https://github.com/usgs/geomag-algorithms/blob/master/geomagio/WebService.py#L292

geomagio/webservice/data.py
```diff
@@ -0,0 +1,351 @@
...
+
+    WebServiceQuery
+        parsed query object
+
+    Returns
+    -------
+    unicode
+        IAGA2002 or JSON formatted string.
+    """
+    if query.output_format == "json":
```
jmfee-usgs: Eliminate the output variables and just return response objects:
```
if query.output_format == "json":
    return Response(
        IMFJSONWriter.format(timeseries, query.elements),
        mimetype="application/json")
else:
    ....
```

geomagio/webservice/data.py
```diff
@@ -0,0 +1,351 @@
...
+        return json_output
+
+    else:
+        iaga_output = IAGA2002Writer.format(timeseries, query.elements)
+        iaga_output = Response(iaga_output, mimetype="text/plain")
+
+        return iaga_output
+
+
+def get_interval(query):
```
jmfee-usgs: Add a `get_interval_from_delta` method in TimeseriesUtility to handle this, similar to `get_delta_from_interval` ( https://github.com/usgs/geomag-algorithms/blob/master/geomagio/TimeseriesUtility.py#L56 ).  All the sampling periods handled by these methods should be considered valid (even though data doesn't exist yet)

geomagio/webservice/data.py
```diff
@@ -0,0 +1,351 @@
...
+                " Valid values are: %s" % (self.sampling_period, ', '.join(VALID_SAMPLING_PERIODS) + '.')
+            )
+        if self.output_format not in VALID_OUTPUT_FORMATS:
+            raise WebServiceException(
+                'Bad format value "%s".'
+                " Valid values are: %s" % (self.output_format, ', '.join(VALID_OUTPUT_FORMATS) + '.')
+            )
+
+
+@blueprint.route("/data", methods=["GET"])
```
jmfee-usgs: Move this method right after `init_app`, since it's an entry point.  All the other methods and classes can be alphabetical after the flask stuff.

geomagio/webservice/data.py
jmfee-usgs: Add a `get_interval_from_delta` method in TimeseriesUtility to handle this, similar to `get_delta_from_interval` ( https://github.com/usgs/geomag-algorithms/blob/master/geomagio/TimeseriesUtility.py#L56 ).  All the sampling periods handled by these methods should be considered valid (even though data doesn't exist yet)

geomagio/webservice/data.py
```diff
@@ -0,0 +1,351 @@
...
+    try:
+        timeseries = get_timeseries(parsed_query)
+        return format_timeseries(timeseries, parsed_query)
+    except Exception as e:
+        message = str(e)
+        error_body = error(500, message, parsed_query, request)
+        return error_body
+
+
+def error(code, message, query, url):
```
jmfee-usgs: rename this to format_error, this should take the status code, exception, and parsed_query objects.  It can check the query format and format as json/plain text

geomagio/webservice/data.py
jmfee-usgs: Add a `get_interval_from_delta` method in TimeseriesUtility to handle this, similar to `get_delta_from_interval` ( https://github.com/usgs/geomag-algorithms/blob/master/geomagio/TimeseriesUtility.py#L56 ).  All the sampling periods handled by these methods should be considered valid (even though data doesn't exist yet)

geomagio/webservice/metadata.py
```diff
@@ -41,8 +41,13 @@ class Metadata(db.Model):
         comment = 'automatic adjusted matrix',
         priority = 1,
         value = {
+<<<<<<< HEAD
```
jmfee-usgs: There's a merge conflict...

geomagio/webservice/data.py
```diff
@@ -0,0 +1,351 @@
...
+"GUA", "SJT"]
+VALID_OUTPUT_FORMATS = ["iaga2002", "json"]
+VALID_SAMPLING_PERIODS = ["1", "60"]
+
+
+blueprint = Blueprint("data", __name__)
+edge_factory = EdgeFactory(
+    host=os.getenv('host', 'cwbpub.cr.usgs.gov'),
+    port=os.getenv('port', 2060),
+    write_port=os.getenv('write_port', 7981)
```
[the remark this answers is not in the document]
trivers-usgs: Updated this to have a `get_input_factory` method that is used within `get_timeseries` to determine the factory based upon the DATA_TYPE environment variable.  Not sure this is the best implementation, but it is working.

geomagio/webservice/data.py
```diff
@@ -0,0 +1,346 @@
...
+        self.endtime = endtime
+        self.elements = elements
+        self.sampling_period = sampling_period
+        self.data_type = data_type
+        self.output_format = output_format
+
+
+def format_error(status_code, exception, query, url):
+    """Assign error_body value based on error format."""
+    error_body = http_error(status_code, exception, query, url)
```
jmfee-usgs: This should do what http_error is doing, and work like format_timeseries and check the query output_format

geomagio/webservice/data.py
```diff
@@ -0,0 +1,346 @@
...
+
+def get_input_factory():
+    """Reads environment variable to determine the factory to be used
+
+    Returns
+    -------
+    input_factory
+        Edge or miniseed factory object
+    """
+    DATA_TYPE = os.getenv('Type', 'edge')
```
jmfee-usgs: The environment variable name should be `DATA_TYPE`, but the variable name should be `data_type`.  Read type, host, and port.

Watch indentation for parameters.

geomagio/webservice/data.py
```diff
@@ -0,0 +1,346 @@
...
+     WebServiceQuery
+        parsed query object
+
+    Returns
+    -------
+    obspy.core.Stream
+        timeseries object with requested data
+    """
+    query.sampling_period = get_interval_from_delta(query.sampling_period)
+    input_factory = get_input_factory()
```
jmfee-usgs: Make input_factory a global, next to blueprint

geomagio/webservice/data.py
```diff
@@ -0,0 +1,346 @@
...
+    """
+    if query.output_format == 'json':
+        http_error_body = json_error(code, message, request.url)
+        return http_error_body
+    else:
+        http_error_body = iaga2002_error(code, message, request.query_string)
+        return http_error_body
+
+
+def iaga2002_error(code, message, request_args):
```
jmfee-usgs: This should just format the iaga2002 error message.

Also, use a multiline string so it's clearer to see the formatting:
```
# code = 400
# status_message = 'Bad Request'
# message = 'Some error description'
# request is the flask request object...
# (Add VERSION to the top as a global, we can set it to a better value later)

return f"""Error {code}: {status_message}

{message}

Usage details are available from {request.base_url}

Request:
{request.url}

Request Submitted
{UTCDateTime().isoformat()}Z

Service Version
{VERSION}
"""
```

geomagio/webservice/data.py
```diff
@@ -0,0 +1,346 @@
...
+    + 'http://geomag.usgs.gov/ws/edge/ \n\n'\
+    + 'Request:\n' + 'ws/edge/?' + str(request_args)[2:-1] + '\n\n' + 'Request Submitted:\n'\
+    + datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ") + '\n'
+
+    error_body = Response(error_body, mimetype="text/plain")
+    return error_body
+
+
+def json_error(code, message, url):
+    """Format json error message.
```
jmfee-usgs: This should just format the json error content.  It would be much more readable using a regular dictionary:

```
error_dict =  {
    "type": "Error",
    "metadata": {
        "status": code,
        # ...
    }
}
return json.dumps(error_dict, sort_keys=True).encode("utf8")
```

geomagio/webservice/data.py
```diff
@@ -0,0 +1,346 @@
...
+
+    Raises
+    ------
+    WebServiceException
+        if any parameters are not supported.
+    """
+    # Create web service query object
+    params = WebServiceQuery()
+
+    # Get and assign values
```
jmfee-usgs: This is still different than before.  It _needs_ to work the same as before, just using flask for argument parsing instead of cgi/http.

This code should remain identical: https://github.com/usgs/geomag-algorithms/blob/master/geomagio/WebService.py#L281-L337 .

And only change the "Get Values" section ( https://github.com/usgs/geomag-algorithms/blob/master/geomagio/WebService.py#L273-L280 )

geomagio/webservice/data.py
trivers-usgs: The factories are still a bit confusing to me, but I made the below changes.  Thoughts?

` data_type = os.getenv('DATA_TYPE', 'edge')

    if data_type == 'edge':
        input_factory = EdgeFactory(
        host=os.getenv('DATA_HOST', 'cwbpub.cr.usgs.gov'),
        port=os.getenv('DATA_PORT', 2060),
        type=data_type
        )
        return input_factory`

trivers-usgs: The factories are still a bit confusing to me, but I made the below changes.  Thoughts?

```
data_type = os.getenv('DATA_TYPE', 'edge')
     if data_type == 'edge':
        input_factory = EdgeFactory(
        host=os.getenv('DATA_HOST', 'cwbpub.cr.usgs.gov'),
        port=os.getenv('DATA_PORT', 2060),
        type=data_type
        )
        return input_factory
```

geomagio/webservice/data.py
trivers-usgs: So, if I use the identical code from the previous webservice, any time it throws errors within the parse method it breaks and doesn't return the nicely formatted errors about invalid parameters.

Also, turns out get() takes in a default parameter if no value is found so I can clean up this section of the parsing a little bit.  

Last - I saw your comment about removing the migrations but wasn't sure how to actually go about that.  Could you show me where?

geomagio/webservice/data.py
```diff
@@ -0,0 +1,372 @@
...
+    # Get values
+    observatory_id = query.get("observatory")
+    start_time = UTCDateTime(query.get("starttime"))
+    end_time = UTCDateTime(query.get("endtime"))
+    elements = query.get("channels", DEFAULT_ELEMENTS)
+    sampling_period = query.get("sampling_period", DEFAULT_SAMPLING_PERIOD)
+    data_type = query.get("type", DEFAULT_DATA_TYPE)
+    output_format = query.get("format", DEFAULT_OUTPUT_FORMAT)
+    # Assign values or defaults
+    if not output_format:
```
jmfee-usgs: If the output format default is set above, I don't think you need this check here.

geomagio/webservice/data.py
```diff
@@ -0,0 +1,372 @@
...
+        data_type = DEFAULT_DATA_TYPE
+    else:
+        data_type = data_type.lower()
+    # Create WebServiceQuery object and set properties
+    params = WebServiceQuery()
+    params.observatory_id = observatory_id
+    params.starttime = start_time
+    params.endtime = end_time
+    params.elements = elements
+    params.sampling_period = str(get_interval_from_delta(sampling_period))
```
jmfee-usgs: interval is already a str, right?

geomagio/webservice/data.py
```diff
@@ -0,0 +1,372 @@
...
+    data_type = query.get("type", DEFAULT_DATA_TYPE)
+    output_format = query.get("format", DEFAULT_OUTPUT_FORMAT)
+    # Assign values or defaults
+    if not output_format:
+        output_format = DEFAULT_OUTPUT_FORMAT
+    else:
+        output_format = output_format.lower()
+
+    observatory_id = observatory_id.upper()
+    if observatory_id not in VALID_OBSERVATORIES:
```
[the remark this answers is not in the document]
trivers-usgs: Agreed, I had moved it out because I was attempting to keep the parse method the same as it is in the old web service.  I'll move it.

geomagio/webservice/data.py
```diff
@@ -0,0 +1,359 @@
...
+    data_type = query.get("type", DEFAULT_DATA_TYPE)
+    output_format = query.get("format", DEFAULT_OUTPUT_FORMAT)
+    # Format values and get time values
+    output_format.lower()
+    observatory_id.upper()
+
+    try:
+        start_time = UTCDateTime(query.get('starttime'))
+    except:
+        start_time = query.get('starttime')
```
jmfee-usgs: It's okay to throw the WebServiceException in parse if there is a _parsing_ error.

If it makes it simpler, you can merge parse and validate.

geomagio/webservice/data.py
```diff
@@ -0,0 +1,359 @@
...
+    if not start_time:
+        now = datetime.now()
+        today = UTCDateTime(
+                year=now.year,
+                month=now.month,
+                day=now.day,
+                hour=0)
+        start_time = today
```
jmfee-usgs: No need for `today` variable, just assign to start_time

geomagio/webservice/data.py
```diff
@@ -0,0 +1,359 @@
...
+
+    try:
+        end_time = UTCDateTime(query.get("endtime"))
+    except:
+        end_time = query.get("endtime")
+    if not end_time:
+        end_time = start_time + (24 * 60 * 60 - 1)
```
jmfee-usgs: UTCDateTime + delta already returns a UTCDateTime

geomagio/webservice/data.py
```diff
@@ -0,0 +1,359 @@
...
+    ----------
+    query: Immutable Dict
+        request.args object
+
+    Raises
+    ------
+    WebServiceException
+        if any parameters are not supported.
+    """
+    if type(query.endtime) == str:
```
jmfee-usgs: If you kept these, I would recommend using `isinstance(query.endtime, str)`, however this should be handled in the `except` portion of parse when the parse error occurs.

geomagio/webservice/data.py
```diff
@@ -0,0 +1,359 @@
...
+    error_dict = {
+        "type": "Error",
+        "metadata": {
+            "status": code,
+            "generated": date,
+            "url": url,
+            "title": status_message,
+            "error": message
+        }
+
```
jmfee-usgs: Watch unneeded blank lines

geomagio/webservice/data.py
```diff
@@ -0,0 +1,359 @@
...
+        parsed query object
+
+    Raises
+    ------
+    WebServiceException
+        if any parameters are not supported.
+    """
+    # Get values
+    observatory_id = query.get("observatory")
+    elements = query.get("channels", DEFAULT_ELEMENTS)
```
jmfee-usgs: We talked about supporting `getlist` for elements, and only splitting if the list has 1 element with commas?

geomagio/webservice/templates/data/usage.html
```diff
@@ -0,0 +1,57 @@
+{% extends '_template.html' %}
+{% block header %}
+<link rel="stylesheet" type="text/css" href="{{ url_for('static', filename='usage.css') }}"/>
+{% endblock %}
+{% block content %}
+  <div class="views-field views-field-field-intro">        <div class="field-content lead pane-border"><p><strong>Example Requests</strong><br /></p><dl><dt>BOU observatory data for current UTC day in IAGA2002 format</dt>
```
jmfee-usgs: Not sure where the `views-field` and other classes come from?  I recommend using semantic elements like `h2` for page sections.  Also, the formatting makes it difficult to scan the content of this template.  We'll have to discuss formatting.

trivers-usgs: I stole this from the current web service documentation page.  It's a mess, but I've cleaned it up a bit now.  I'm all for another formatter.

jmfee-usgs:
This was merged externally, with the following additional fixes https://github.com/usgs/geomag-algorithms/commit/55b6c37f3b3168e56f222330d82aacf0f182838d

PR #275, opened 2020-02-20
WIP: Initial attempt at residual api and factory

jmfee-usgs:
@pcain-usgs, @aclaycomb-usgs  :  based on discussion the other day, and portions of the existing adjusted notebook.

What are your thoughts?  I think some changes are needed:
- Other parts of the library use radians internally
- Maybe some reading attributes should be more explicit like pier_correction

Eventually would add a "residual spreadsheet factory" to read existing residual method measurements and calculated absolutes that could be used for validation...


geomagio/residual/WebAbsolutesFactory.py
```diff
@@ -0,0 +1,83 @@
...
+            "reviewer": data["reviewer"],
+        }
+
+    def _parse_reading(self, metadata, data, include_measurements):
+        absolutes = [
+            self._parse_absolute(element, data[element])
+            for element in ["D", "H", "Z"]
+            if element in data
+        ]
+        measurements = (
```
erigler-usgs: Just curious, why is this a tuple instead of a list?

jmfee-usgs: This is just a grouping for the statements that follow, from the black auto formatter, and measurements is assigned either the parse_measurement list comprehension or an empty list...

erigler-usgs: Ah. I thought it was some kind of "tuple comprehension". Then I searched this on Google, and learned that the only way to do *that* is `tuple(item for item in iterable)`, since parentheses are already used for a generator comprehension. Thanks, and resolved.

pcain-usgs:
> decimal degrees, or decimal radians. I think the calculations use decimals... formatting as DMS or parsing from DMS is not a big deal.

The calculations, for clarification, use decimal radians.

aclaycomb-usgs:
would this be where the projects from bill, john etc would all be implemented, including replacing make_cal?

pcain-usgs:
Of the metadata included in the original script, I believe that the following keys are relevant to calculating absolutes:

- Azimuth Angle
- Pier Correction
- Hemisphere: 1 or -1 depending on northern or southern hemisphere
- Starttime
- Endtime

It appears that the rest of the relevant information for this task is calculated on the fly.

jmfee-usgs:
@pcain-usgs I added key metadata to the Reading class and some other updates to split each residual into a separate object.  Let's plan to discuss next week.  Here is an updated example to fetch an observation:

```
from obspy import UTCDateTime
import json
from geomagio.residual.WebAbsolutesFactory import WebAbsolutesFactory

factory = WebAbsolutesFactory()
readings = factory.get_readings(
    "BOU", UTCDateTime("2020-01-03"), UTCDateTime("2020-01-04"), True
)

def json_default(obj):
    return isinstance(obj, UTCDateTime) and obj.isoformat() or obj.__dict__

for r in readings:
    print(json.dumps(r.absolutes_index(), default=json_default, indent=2))
    print(json.dumps(r.measurement_index(), default=json_default, indent=2))
```

jmfee-usgs:
Incorporated #289

PR #276, opened 2020-02-20
Make filter algorithm process any channels

### LINKED ISSUE #274: FilterAlgorithm quietly fails if all specified input channels are not available
If you specify (for example) `--inchannels H E Z F` for the Controller, and `F` is empty, the Filter algorithm will not return anything, even though H, E, and Z are perfectly good inputs.

The reason for this is that the default `can_produce_data` method for the parent Algorithm class that all specified `--inchannels` be available in order to return TRUE, otherwise it will not attempt to process any further

https://github.com/usgs/geomag-algorithms/blob/1c06af9d0a2e00b426da09c6b10417cc2ce9398c/geomagio/Controller.py#L268-L272

The fix should probably be for a `can_produce_data` method in FilterAlgorithm.py to return `True` if there are *any* traces in the passed stream. However, GitHub was kind enough to point out that this Issue is similar to #205, where I once proposed a similar "fix", but a slightly different approach was taken in the end. I would like to ping @jmfee-usgs to look at this briefly and weigh in with an opinion.

jmfee-usgs:
Fixes #274 (I think).

@erigler-usgs take a look, this makes filter run if any of its input channels have data.  It does not check whether there are enough samples to produce valid data, or non-nan output.  I think this is along the lines of what you described.

It may be possible to apply this to other algorithms that have variable input requirements, but that is left for later.

### DIFF
--- a/geomagio/TimeseriesUtility.py
+++ b/geomagio/TimeseriesUtility.py
@@ -270,8 +270,6 @@ def has_all_channels(stream, channels, starttime, endtime):
     """
     input_gaps = get_merged_gaps(
             get_stream_gaps(stream=stream, channels=channels))
-    print(starttime, endtime)
-    print(input_gaps)
     for input_gap in input_gaps:
         # Check for gaps that include the entire range
         if (starttime >= input_gap[0] and
--- a/geomagio/TimeseriesUtility.py
+++ b/geomagio/TimeseriesUtility.py
@@ -250,6 +250,74 @@ def get_channels(stream):
     return [ch for ch in channels]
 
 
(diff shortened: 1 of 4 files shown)


erigler-usgs:
I see you did not actually implement a "channel-foreach" option (as the name of this PR might imply), but essentially did what I originally proposed, albeit in a more elegant way. Are you comfortable with this approach after reading my follow-up comments in #274 about updating data in channel A when there is only a gap in channel B? If so, I'm happy to merge this PR.

erigler-usgs:
I'm going to merge because I think the added functionality will be generally useful, even if it is not used in an eventual "channel-foreach" option.

PR #280, opened 2020-03-04
Develop docs

jmfee-usgs:
@erigler-usgs : anything else this should cover?


jmfee-usgs:
`obspy` lists numpy, scipy, and matplotlib as dependencies, so pip should install those too.

PR #283, opened 2020-03-10
Use setuptools-pipfile so dependencies managed in pipfile

jmfee-usgs:
- use pipfile for pip dependencies
- support ssl intercept configuration if SSL_CERT_FILE environment configured
- move pre-commit to dev dependencies

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,11 +39,12 @@ RUN useradd \
         -r \
         -s /sbin/nologin \
          geomag_user \
-    && mkdir -p /home/geomag_user/notebooks \
-    && chown -R geomag_user:geomag_user /home/geomag_user
+    && mkdir -p /data \
+    && chown -R geomag_user:geomag_user /data
 
 USER geomag_user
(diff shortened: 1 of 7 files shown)


jmfee-usgs:
@emartinez-usgs this seems to work for setuptools-pipfile

erigler-usgs:
@jmfee-usgs , I think this looks good.

Even though it may not necessarily belong with this PR, I will ask if it makes sense to update `docs/install.md` to say something about pipenv. I know it is described in `docs/develop.md`, but shouldn't we have something that describes how non-developers might use it, kinda like we already do for conda in `docs/install.md`?

jmfee-usgs:
@erigler-usgs I made a few changes in response to your comments.

erigler-usgs:
Thanks. Your changes aren't quite what I was expecting, so I suspect I still lack some understanding of how pipenv is supposed to be used. I'll approve, merge, and ask you to re-educate me at a later date.

Issue #281, opened 2020-03-06
Split flask web service into 2 flask apps


jmfee-usgs:
Currently, flask sessions use the database even when making data requests.

This article describes a solution where different flask apps are loaded based on different urls:
https://www.peterspython.com/en/blog/two-flask-apps-frontend-and-admin-on-one-domain-using-dispatchermiddleware

I suggest splitting app.py into "public_app.py" and "restricted_app.py".  The public app will include the data service for now.  The internal app will include the sessions and auth.

"/ws" -> public
"/restricted/ws" -> restricted

jmfee-usgs:
@trivers88 : this is the issue I mentioned earlier

PR #285, opened 2020-03-16
Split web service into two separate apps

trivers-usgs:
Addresses #281


trivers-usgs:
Could you point out what I'm missing when you have a chance @jmfee-usgs ?

geomagio/webservice/run_both.py
```diff
@@ -0,0 +1,22 @@
...
+from werkzeug.serving import run_simple
+
+from .public_app import create_app as create_public_app
+from .restricted_app import create_app as create_restricted_app
+
+public_app = create_public_app()
+restricted_app = create_restricted_app()
+
+# merge
+application = DispatcherMiddleware(public_app, {"/restricted": restricted_app})
```
jmfee-usgs: Make this a `create_app` function that returns the DispatcherMiddleware (and create the other apps too).  It will work more similar to the existing apps.

Rename to `app.py` since it's what we plan to run. (and update `__init__.py` to be how it was before)

geomagio/webservice/restricted_app.py
```diff
@@ -0,0 +1,27 @@
...
+
+def create_app():
+    app = flask.Flask(__name__, instance_relative_config=True)
+    # configure using environment variables
+    app.config.update(os.environ)
+
+    # connect modules
+    database.init_app(app)
+    login.init_app(app)
+    data.init_app(app)
```
jmfee-usgs: The restricted app should have sessions, but not data.

geomagio/webservice/public_app.py
```diff
@@ -6,18 +6,16 @@
...
 
 def create_app():
-    app = flask.Flask(__name__)
+    app = flask.Flask(__name__, instance_relative_config=True)
     # configure using environment variables
     app.config.update(os.environ)
 
     # connect modules
     database.init_app(app)
     login.init_app(app)
```
jmfee-usgs: The public app should only have `data`, and no database, login, or session.

geomagio/webservice/run_public.py
```diff
@@ -0,0 +1,6 @@
+from .public_app import create_app
+
+public_app = create_app()
+
+if __name__ == "__main__":
```
jmfee-usgs: Move this (same for restricted) into `public_app.py`.  No need for a separate "run" file.

jmfee-usgs: This may also not be needed, since I think you can just point to the individual files when running flask (i.e. `flask run geomagio.webservice.public_app`)

jmfee-usgs:
Not sure why the auto formatter didn't run.  Make sure you have `pre-commit` installed locally by running:

`pre-commit install`

geomagio/webservice/public_app.py
jmfee-usgs: ^ still shouldn't have `login` in the public app

PR #287, opened 2020-03-23
Fast Api web service

trivers-usgs:


geomagio/API/data_api.py
```diff
@@ -0,0 +1,222 @@
...
+        parsed_query = parse_query(observatory)
+        validate_query(parsed_query)
+    except Exception as e:
+        return format_error(400, e)
+
+    return parsed_query
+
+
+def format_error(status_code, exception):
+    return JSONResponse(json_error(status_code, exception), mimetype="application/json")
```
jmfee-usgs: I think JSONResponse isn't needed, and you can just return a `Dict` from json_error.  Fastapi will automatically convert to json

geomagio/API/data_api.py
```diff
@@ -0,0 +1,222 @@
...
+
+@app.get("/")
+def read_root():
+    return {"Hello": "World"}
+
+
+@app.get("/data/")
+async def read_query(
+    id: str,
+    starttime: Any = Query(UTCDateTime(datetime.now())),
```
jmfee-usgs: I'm not sure this works as intended, I think it sets the default starttime to the time the web service starts.

Make the type datetime.datetime, and the default None, then check for the default and use the old "start of current day" logic.

geomagio/API/data_api.py
```diff
@@ -0,0 +1,222 @@
...
+    endtime: str = Query(None),
+    elements: List[str] = Query(DEFAULT_ELEMENTS),
+    sampling_period: float = DEFAULT_SAMPLING_PERIOD,
```
jmfee-usgs: I think if you create an enumeration, fastapi will handle validation for you

```
import enum

class SamplingPeriod(float, enum.Enum):
    TEN_HERTZ = 0.1
    SECOND = 1.0
    # ...
```

jmfee-usgs: Same for data_type and output_format, although data_type may need some extra work to support location codes.  You could try something like `Union[DataType, str]` and then double check that str is 2 characters if it's not a known DataType.

geomagio/API/data_api.py
```diff
@@ -0,0 +1,222 @@
...
+    output_format: str = DEFAULT_OUTPUT_FORMAT
+
+
+@app.get("/")
+def read_root():
+    return {"Hello": "World"}
+
+
+@app.get("/data/")
+async def read_query(
```
jmfee-usgs: You might check out `Depends`.  I think you could use it to parse the query object and pass into a `get_data` endpoint.  If that's too weird, just merge parsing into the get_data method.

geomagio/API/data_api.py
```diff
@@ -0,0 +1,332 @@
...
+    observatory_id: str
+    starttime: Any
+    endtime: Any
+    elements: List[str]
+    sampling_period: SamplingPeriod
+    data_type: Union[DataType, str]
+    output_format: OutputFormat
+
+
+@app.get("/data/")
```
jmfee-usgs: Keep the global `app` variable and this method together after the constants.  Then classes, and then methods, each in alphabetical order.

geomagio/API/data_api.py
```diff
@@ -0,0 +1,332 @@
...
+            "status": code,
+            "error": str(error),
+            "generated": UTCDateTime().isoformat() + "Z",
+            "url": url,
+        },
+    }
+    return dumps(error_dict).encode("utf8")
+
+
+def validate_query(query):
```
jmfee-usgs: Convert validate to be a pydantic validator method.  You could validate '*' to check all fields at once
https://pydantic-docs.helpmanual.io/usage/validators/

It's probably worth moving this into a separate file that data_api can import.  Rename WebServiceQuery to DataApiQuery or similar, since we'll have more endpoints soon.

geomagio/api/data/data_api.py
```diff
@@ -279,14 +161,14 @@ def json_error(code: int, error: Exception, url):
...
-    sampling_period: SamplingPeriod = Query(DEFAULT_SAMPLING_PERIOD),
-    data_type: Union[DataType, str] = Query(DEFAULT_DATA_TYPE),
-    format: OutputFormat = Query(DEFAULT_OUTPUT_FORMAT),
-) -> WebServiceQuery:
+    sampling_period: SamplingPeriod = Query(SamplingPeriod.HOUR),
+    data_type: Union[DataType, str] = Query(DataType.VARIATION),
+    format: OutputFormat = Query(OutputFormat.IAGA2002),
+) -> DataApiQuery:
 
     if len(elements) == 0:
```
jmfee-usgs: Move this stuff into validator methods, using always=True to set defaults

geomagio/api/data/data_api.py
```diff
@@ -35,122 +27,12 @@
...
-    "SJG",
-    "SJT",
-    "TST",
-    "TUC",
-    "USGS",
-]
+
 VERSION = "version"
 
 app = FastAPI(docs_url="/data")
```
jmfee-usgs: Move `app` to the end, by the methods that use it.

geomagio/api/data/data_api.py
```diff
@@ -1,31 +1,23 @@
...
-from ..TimeseriesUtility import get_interval_from_delta
-from fastapi.exception_handlers import (
-    http_exception_handler,
-    request_validation_exception_handler,
-)
+from ...edge import EdgeFactory
+from ...iaga2002 import IAGA2002Writer
+from ...imfjson import IMFJSONWriter
+from ...TimeseriesUtility import get_interval_from_delta
+from .DataApiQuery import *
```
jmfee-usgs: list all the classes even if all of them are imported

geomagio/api/data/data_api.py
```diff
@@ -1,31 +1,23 @@
...
-from typing import List, Any, Union
+from typing import Any, List, Union
 
-from fastapi import FastAPI, Query, Request, Depends, HTTPException
+from fastapi import Depends, FastAPI, Query, Request
 from obspy import UTCDateTime
-from pydantic import BaseModel, validator, ValidationError, root_validator
-from starlette.responses import PlainTextResponse, Response
-from fastapi.encoders import jsonable_encoder
 from fastapi.exceptions import RequestValidationError
```
jmfee-usgs: keep all the fastapi imports together

geomagio/api/data/data_api.py
```diff
@@ -4,20 +4,19 @@
...
 
 from fastapi import Depends, FastAPI, Query, Request
-from obspy import UTCDateTime
 from fastapi.exceptions import RequestValidationError
 from fastapi.exception_handlers import request_validation_exception_handler
 from fastapi.responses import JSONResponse
+from obspy import UTCDateTime
 from starlette.responses import Response
 
+from .DataApiQuery import DataApiQuery, DataType, OutputFormat, SamplingPeriod
```
jmfee-usgs: Imports that are closer (same package instead of other outside packages) should be closer to the code (at the end of the imports list)

geomagio/api/data/data_api.py
```diff
@@ -30,10 +29,11 @@
 
 VERSION = "version"
 
-app = FastAPI(docs_url="/data")
-
 
-def format_error(status_code, exception, format, request):
+def format_error(
+    status_code: int, exception: str, format: str, request: Request
+) -> str or dict:
```
jmfee-usgs: This should probably return `Union[Dict, Response]` or just `Response` if you don't return json_error directly...

geomagio/api/data/data_api.py
```diff
@@ -46,15 +46,15 @@ def format_error(status_code, exception, format, request):
     return error
 
 
-def format_timeseries(timeseries, query):
+def format_timeseries(timeseries, query: DataApiQuery) -> str:
```
jmfee-usgs: timeseries will be a `Stream` (from obspy)

geomagio/api/data/data_api.py
```diff
@@ -138,7 +145,7 @@ def iaga2002_error(code, error, url):
     return error_body
 
 
-def json_error(code: int, error: Exception, url):
+def json_error(code: int, error: Exception, url: str) -> dict:
```
jmfee-usgs: This should be `Dict` to refer to the type

geomagio/api/data/data_api.py
```diff
@@ -162,18 +169,17 @@ def json_error(code: int, error: Exception, url):
...
     id: str,
-    starttime: Any = Query(None),
-    endtime: Any = Query(None),
-    elements: List[str] = Query(DEFAULT_ELEMENTS),
-    sampling_period: SamplingPeriod = Query(SamplingPeriod.HOUR),
-    data_type: Union[DataType, str] = Query(DataType.VARIATION),
-    format: OutputFormat = Query(OutputFormat.IAGA2002),
+    starttime: datetime = Query(None),
+    endtime: datetime = Query(None),
+    elements: List[str] = Query(None),
```
jmfee-usgs: You might want to keep the defaults here so they are mentioned in API docs.

Setting them in DataApiQuery is helpful for other parts of the library that might create a DataApiQuery object...

geomagio/api/data/data_api.py
```diff
@@ -162,18 +169,17 @@ def json_error(code: int, error: Exception, url):
...
+    sampling_period: SamplingPeriod = Query(None),
+    data_type: Union[DataType, str] = Query(None),
+    format: OutputFormat = Query(None),
 ) -> DataApiQuery:
 
-    if len(elements) == 0:
-        elements = DEFAULT_ELEMENTS
-    if len(elements) == 1 and "," in elements[0]:
-        elements = [e.strip() for e in elements[0].split(",")]
+    if elements != None:
```
jmfee-usgs: This can be combined into one if:

```
if elements and len(elements) == 1 ...
```

geomagio/api/data/data_api.py
```diff
@@ -162,18 +169,17 @@ def json_error(code: int, error: Exception, url):
...
+        if len(elements) == 1 and "," in elements[0]:
+            elements = [e.strip() for e in elements[0].split(",")]
 
     if starttime == None:
         now = datetime.now()
```
jmfee-usgs: This can go into the DataApiQuery pre validator to set a default

PR #290, opened 2020-04-06
Add observatories endpoint

trivers-usgs:


jmfee-usgs:
I think your file names are swapped.

`Observatory` - should be capitalized because it is a class.
`observatories` - should be lowercased because it is a route.

I probably need to clean up /ws/secure stuff to match...

trivers-usgs:
I literally had the file names reversed but changed them right before I pushed for no reason

PR #292, opened 2020-04-10
WIP: Residual api

jmfee-usgs:
@pcain-usgs : creating pull request so I can leave comments.  When you feel this is ready to merge, remove the `WIP: ` prefix and let me know.


geomagio/residual/Calculation.py
```diff
@@ -0,0 +1,399 @@
...
+    shift: Degree shift in inclination measurements.
+    """
+
+    def __init__(
+        self,
+        angle: float = None,
+        residual: float = None,
+        ordinate: Ordinate = None,
+        f: float = None,
+        ud: int = None,
```
jmfee-usgs, line 26: Update this to extend from pydantic `BaseModel`.

Use more descriptive attribute names, like `up_down` or `direction` instead of `ud`.  You could also create an enumeration like this when you expect specific values:
```
import enum
class MeasurementDirection(int, enum.Enum):
  UP = 1
  DOWN = -1
```

geomagio/residual/Calculation.py
```diff
@@ -0,0 +1,399 @@
...
+
+def calculate(Reading):
+    """
+    Calculate/recalculate absolute from a Reading object's
+    ordinates, measurements, and metadata.
+    Outputs a list of absolutes containing baseline, absolute,
+    and element name. Also reutrns the scale value.
+    """
+    # define measurement types used to calculate inclination
+    inclination_types = [mt.NORTH_DOWN, mt.NORTH_UP, mt.SOUTH_DOWN, mt.SOUTH_UP]
```
jmfee-usgs, line 47: Make this either a global (`INCLINATION_TYPES`) defined at the top, or inline into the list comprehension:

```
[
  o for o in Reading.ordinates
  if o.measurement_type in [ ... ]
]
```

geomagio/residual/Calculation.py
```diff
@@ -0,0 +1,399 @@
...
+        Reading.ordinate_index(),
+        mean,
+        Reading.metadata,
+    )
+    # calculate absolutes
+    Habs, Zabs = calculate_absolutes(f, inclination)
+    # calculate baselines for H and Z
+    Hb, Zb = calculate_baselines(Habs, Zabs, mean, Reading.pier_correction)
+    # calculate scale value
+    scale_ordinates = Reading.ordinate_index()["NorthDownScale"]
```
jmfee-usgs, line 67: Use the enumeration values instead of strings (`mt.NorthDownScale`)

geomagio/residual/Calculation.py
```diff
@@ -0,0 +1,399 @@
...
+        self.angle = angle
+        self.residual = residual
+        self.ordinate = ordinate
+        self.f = f
+        self.ud = ud
+        self.shift = shift
+        self.pm = pm
+
+
+def calculate(Reading):
```
jmfee-usgs, line 39: Use lower case variable names, so this isn't confused with the class `Reading`

geomagio/residual/Calculation.py
```diff
@@ -0,0 +1,399 @@
...
+    inclination_types = [mt.NORTH_DOWN, mt.NORTH_UP, mt.SOUTH_DOWN, mt.SOUTH_UP]
+    # get ordinate values across h, e, z, and f for inclination measurement types
+    inclination_ordinates = [
+        o for o in Reading.ordinates if o.measurement_type in inclination_types
+    ]
+    # get average ordinate values across h, e, z, and f
+    mean = average_ordinate(inclination_ordinates, None)
+    # calculate inclination
+    inclination, f = calculate_I(
+        Reading.measurement_index(),
```
jmfee-usgs, line 56: make variables for the indices at the top of the function, since they are used multiple times:
```
measurement_index = reading.measurement_index()
```

geomagio/residual/Calculation.py
```diff
@@ -0,0 +1,399 @@
...
+        inclination_ordinates,
+        Reading.ordinate_index(),
+        mean,
+        Reading.metadata,
+    )
+    # calculate absolutes
+    Habs, Zabs = calculate_absolutes(f, inclination)
```
jmfee-usgs, line 63: Use underscores in variable names (`h_abs`, `z_abs`) where possible.  Title case is used for class names.

geomagio/residual/Calculation.py
```diff
@@ -0,0 +1,399 @@
...
+    if Iprime >= 100:
+        Iprime -= 200
+    # get multiplier for hempisphere the observatory is located in
+    # 1 if observatory is in northern hemisphere
+    # -1 if observatory is in southern hemisphere
+    hs = metadata["hemisphere"]
+    # gather calculation objects for each measurement type
+    southdown = Calculate(
+        shift=-200,
+        ud=1,
```
jmfee-usgs, line 111: I think you can handle `shift`, `direction` (aka ud), and `meridian` (?aka pm) as part of the MeasurementType enum to make this much easier to read.

Add properties to the end of the MeasurementType class that return these values based on the values:
```
    @property
    def direction(self):
        if self in [MeasurementType.SOUTH_DOWN, MeasurementType.NORTH_DOWN]:
            return 1
        else:
            return -1
```

Then in the code you can do things like:
```
mt.SOUTH_DOWN.direction
```

jmfee-usgs: Then you could simplify this into a list comprehension:

```
values = {
  m: Calculate(m.shift, m.direction, m.meridian, angle=average_angle(measurements, m), ...)
  for m in [mt.SOUTH_DOWN, mt.SOUTH_UP, ...]
}
```

`values[mt.SOUTH_DOWN]`, etc

geomagio/residual/SpreadsheetAbsolutesFactory.py
```diff
@@ -42,7 +43,167 @@
...
-    {"type": mt.NORTH_DOWN, "angle": "D45", "residual": "E45", "time": "B45"},
+    # scaling
+    {"type": mt.NORTH_DOWN_SCALE, "angle": "D44", "residual": "E44", "time": "B44"},
+    {"type": mt.NORTH_DOWN_SCALE, "angle": "D45", "residual": "E45", "time": "B45"},
+]
+
+SPREADSHEET_ORDINATES = [
+    # declination
+    {
+        "type": mt.WEST_DOWN,
```
jmfee-usgs, line 54: you might combine the `h`, `e`, `z`, and `f` attributes into SPREADSHEET_MEASUREMENTS, since the times and measurement types are already there...

geomagio/residual/Calculation.py
pcain-usgs: Would you want this difference to propagate across the entire module or only this line?

test/residual_test/Calculation_test.py
```diff
@@ -0,0 +1,42 @@
+from geomagio.residual import SpreadsheetAbsolutesFactory
+from numpy.testing import assert_almost_equal
+
+import os
+
+for filename in os.listdir("etc/residual")[1::]:
```
jmfee-usgs, line 6: put this inside a method called "test_SOMETHING" and pytest will add it to the results...  You might also make this method "read_and_compare_absolutes(file)" and have a separate method for each file that describes why it makes a good test case.

Calculation_test.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+for filename in os.listdir("etc/residual"):
+    if filename == ".DS_Store":
+        continue
+    # establish SpreadsheetAbsolutesFactory for reading test data from Excel
+    saf = SpreadsheetAbsolutesFactory()
+    # Read spreadsheet containing test data
+    reading = saf.parse_spreadsheet("etc/residual/" + filename)
+    # establish original absolute object
```
jmfee-usgs, line 14: Also separate the function that recalculates and compares absolutes, so it can be used for both spreadsheet and web absolute evaluation.

test/residual_test/residual_test.py
```diff
@@ -0,0 +1,100 @@
...
+    Compare calulations to original absolutes obejct from Spreadsheet.
+    Tests gathering of Dedhorse's metadata for use by calculations.
+    Tests calculations for measurements in units of DMS.
+    """
+    # establish SpreadsheetAbsolutesFactory for reading test data from Excel
+    saf = SpreadsheetAbsolutesFactory()
+    # Read spreadsheet containing test data
+    reading = saf.parse_spreadsheet("etc/residual/DED-20140952332.xlsm")
+    # establish original absolute object
+    original = reading.absolutes
```
jmfee-usgs, line 34: I think this part should be a separate method that each test uses, instead of repeating:
```
def check_absolute_calculation(reading: Reading):
    # save existing absolutes
    # calculate absolutes
    # compare
```

Also, update the comparison to use the absolutes_index, so it compares "H" to "H" without relying on the order...  If you want to get fancy, this could be a non-test method that could be used during loading too to flag any absolutes that have changed and might need further review.  It could general a List[str] of warnings, and the test could fail if the list of warnings is non-empty.

jmfee-usgs:
@pcain-usgs : merging for now, I'll request additional changes in a new issue

PR #293, opened 2020-04-14
Switch orm

jmfee-usgs:
@trivers-usgs let me know if this fixes the angular issue


jmfee-usgs:
Sounds like this has fixed the angular redirect problem.

PR #294, opened 2020-04-14
Middleware

trivers-usgs:
Add allow_origins to middleware


jmfee-usgs:
Please rebase...

trivers-usgs:
I broke stuff yesterday trying to add your branch and then rebase.  I think I fixed it this time around though.

jmfee-usgs:
I'm not sure this is the right direction, closing for now.

Both the web service and app will be deployed on the same site in the long run, so for development purposes, I think you should set up a proxy so angular forwards /ws requests to the api.

PR #299, opened 2020-05-06
Controller refactor

jmfee-usgs:
Make run and run_as_update more reusable outside controller.


erigler-usgs:
@jmfee-usgs, I think this sounds like a great idea, but given it's potential to break production processing, I want to look at it very closely, and test exhaustively in a production-like environment. It might take a little more time than usual.

jmfee-usgs:
Sure, I'm working on adding some of the filtering and variation processing steps as well.

The intent is to reduce the number of processes needed, and we can test in the staging environment.  I'll coordinate when ready, but wanted it to be on your radar.

jmfee-usgs:
@erigler-usgs : not ready for merge, but will you take a look at the direction this is going and comment?

geomagio/Controller.py
```diff
@@ -151,54 +63,27 @@ def run(self, options, input_timeseries=None):
...
-            processed = self._rename_channels(
-                timeseries=processed, renames=options.rename_output_channel
-            )
-        # output
-        self._outputFactory.put_timeseries(
-            timeseries=processed,
-            starttime=starttime,
-            endtime=endtime,
-            channels=output_channels,
+        run(
```
erigler-usgs, line 66: I understand that this is not recursive, since run() is not the same as self.run(), but this name might be a little confusing for the uninitiated.

geomagio/processing/run.py
```diff
@@ -0,0 +1,284 @@
+"""Controller class for geomag algorithms"""
```
erigler-usgs, line 1: This is not a "class"? Is the plan for it to evolve into one?

jmfee-usgs: @erigler-usgs : there are options.  not sure what you think of the concurrent processing elsewhere? or if there are other ways you'd like to be able to run these algorithms?

PR #301, opened 2020-05-08
Fixes issue 115

pcain-usgs:
Implementation of a simple DBDT algorithm using second order central differences(np.gradient)


geomagio/algorithm/__init__.py
```diff
@@ -14,6 +14,7 @@
 from .FilterAlgorithm import FilterAlgorithm
 from .SqDistAlgorithm import SqDistAlgorithm
 from .XYZAlgorithm import XYZAlgorithm
+from .DBDTAlgorithm import DBDTAlgorithm
```
jmfee-usgs, line 17: sort alphabetically, unless otherwise required not to (imports and code below)

geomagio/algorithm/DBDTAlgorithm.py
```diff
@@ -0,0 +1,48 @@
...
+        stream : obspy.core.Stream
+            stream of data to process
+        Returns
+        -------
+        out : obspy.core.Stream
+            stream containing 1 trace per original trace.
+        """
+        out = Stream()
+        for trace in stream:
+            dbdt = np.gradient(trace.data)
```
jmfee-usgs, line 32: is `np.diff` more accurate for what is desired, since it is first difference?  gradient seems to be more involved

geomagio/algorithm/DBDTAlgorithm.py
```diff
@@ -0,0 +1,48 @@
...
+            stats = Stats(trace.stats)
+            stats.channel = "DBDT-" + stats.channel
```
jmfee-usgs, line 34: I would make this "{channel}_DDT" or similar.

It would be good to configure this suffix as a location code in MiniSeed Factory, RB, AB, etc?:
https://github.com/usgs/geomag-algorithms/blob/master/geomagio/edge/MiniSeedFactory.py#L351

geomagio/algorithm/DBDTAlgorithm.py
```diff
@@ -0,0 +1,48 @@
...
+            trace_out = create_empty_trace(
+                starttime=stats.starttime,
```
jmfee-usgs, line 36: Depending on diff/gradient, and definition of dB/dt, should there be a one sample time shift in the new trace?  Similarly, should the algorithm request one leading/trailing sample using the get_interval api method?

geomagio/algorithm/DBDTAlgorithm.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import absolute_import
+from ..TimeseriesUtility import create_empty_trace, get_interval_from_delta
+from .Algorithm import Algorithm
+import numpy as np
+from obspy.core import Stream, Stats
+
+
+class DBDTAlgorithm(Algorithm):
```
jmfee-usgs, line 8: Consider making this `DbDtAlgorithm`, since "Db" is one unit, etc.

geomagio/algorithm/DBDTAlgorithm.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import absolute_import
+from ..TimeseriesUtility import create_empty_trace, get_interval_from_delta
+from .Algorithm import Algorithm
+import numpy as np
```
jmfee-usgs, line 4: Group the imports.  I think now that library is 3.6+, we're in the future and __future__ is no longer needed :)

Group by:
- standard library
- external dependencies (numpy, obspy, etc)
- imports from this package (.. before .)

within each group, use alphabetical order

pcain-usgs:
Example usage in terminal environment:

python geomag.py --input edge --observatory BOU --starttime 2020-05-01T00:00:00Z --endtime 2020-05-01T01:00:00Z --type variation --interval second --inchannels H --algorithm dbdt --outchannels H_DDT --output plot &

erigler-usgs:
I'm curious about the change from "DBDT" to "DbDt". Remember, it's calculating the derivative of B with respect to time. If anything, the name should probably be "dBdt", which is how most people write it out (if not dB/dt). I'd be just as happy with all the letters being in the same case, upper or lower. But, as much as I usually like "camel case", this just looks awkward to me.

jmfee-usgs:
Classes start with upper case, and "Db" and "Dt" are each a logical unit.  This is mainly to help read the name.  Comments are are free to use a different notation...

geomagio/edge/MiniSeedFactory.py
```diff
@@ -421,6 +421,8 @@ def _get_edge_location(self, observatory, channel, data_type, interval):
                 location_suffix = "Q"
             elif suffix == "SV":
                 location_suffix = "V"
+            elif suffix == "DDT":
+                location_suffix = "DT"
```
jmfee-usgs, line 425: `location_suffix` can only be 1 character (because combined with prefix it can only be 2 chars total).  `D` is already used for Dist.  Also, not sure whether this should be stored in Edge.  @erigler-usgs : thoughts?

jmfee-usgs:
Rebased and merged externally with fixes.

PR #303, opened 2020-05-14
AdjustedAlgorithm update

pcain-usgs:
Extends functionality from H,E,Z,F conversions(X,Y,Z,F) to user specified inchannels and outchannels.


geomagio/algorithm/AdjustedAlgorithm.py
```diff
@@ -21,24 +21,26 @@ def __init__(
         statefile=None,
         data_type=None,
         location=None,
+        inchannels=None,
+        outchannels=None,
     ):
-        Algorithm.__init__(
-            self, inchannels=("H", "E", "Z", "F"), outchannels=("X", "Y", "Z", "F")
```
jmfee-usgs, line 26: Is there a reason to require these to be specified correctly, instead of using defaults?  Isn't this order important?

geomagio/algorithm/AdjustedAlgorithm.py
```diff
@@ -51,49 +53,34 @@ def load_state(self):
...
-        self.matrix[3, 2] = np.float64(data["M43"])
-        self.matrix[3, 3] = np.float64(data["M44"])
-        self.pier_correction = np.float64(data["PC"])
+
+        PC = data.pop("PC")
+        self.pier_correction = np.float64(PC)
+        # excludes PC
+        keys = list(data.keys())
+        # get maximum row/colum number
+        length = int(max(keys)[-1])
```
jmfee-usgs, line 62: What if the matrix were stored as an array, instead of with individual keys?
```
"pc": x,
"matrix": [1, 2, 3, 4, ...]
```

geomagio/algorithm/AdjustedAlgorithm.py
pcain-usgs: Josh expressed an interest in utilizing this algorithm to adjust our raw electric field data. The same operation can be used to fix an underlying issue in that data, reverse polarity. This edit removes the defaults from HEZF and XYZF and allows a user to specify what channel is being converted/what channel it is specifying as output. Just makes the operation more accessible past its primary use.

jmfee-usgs: Ok, maybe make it configurable to override the defaults, but keep the defaults for existing users...

geomagio/algorithm/AdjustedAlgorithm.py
pcain-usgs: This is an adaptation of how the matrices where originally created in json. I think this would be a more appropriate question for Josh. I know that some of the documentation utilizes the dictionary's keys(M11 ... M44) as a part of its explanation so making this change may require that we update documentation.

geomagio/algorithm/AdjustedAlgorithm.py
erigler-usgs: Just to add to what Payton said, I envision other transformations besides just E-field that will be able to use the basic functionality provided by the Adjusted algorithm. I don't want to be constrained by specific input/output channel names. But I have no issue falling  back to HEZF and XYZF as defaults.

geomagio/algorithm/AdjustedAlgorithm.py
erigler-usgs: It is a little problematic, although certainly not impossible, to serialize 2D matrices into a JSON file. You can do something like `json.dumps(matrix.tolist())`, which will create a list of lists. But, doing `json.dumps(list(array))` does not recursively convert each row of array into a list. Rather, you just get a list of 1D Numpy arrays.

But for now, I recommend not deviating too far from Abe's original implementation, which while a bit ugly, is absolutely clear about where each matrix coefficient is supposed to go, and is completely agnostic as to which programming language or numerical library is used.

erigler-usgs:
@pcain-usgs ,

Can you add a test or two that shows that this works for something less than 4x4 matrix, and maybe something with no PC? Abe already included some test data, maybe you could re-use this, trimming off any unused channel(s) and affine matrix elements as appropriate. Assuming that works as expected, I'm prepared to merge this.

erigler-usgs:
When I run the Adjusted algorithm from the command line now, it handles arbitrary `--inchannels` as intended, but if I do NOT pass anything via `--inchannels`
```
geomag.py --input edge --input-host cwbpub.cr.usgs.gov --output iaga2002 --output-stdout --observatory BOU --algorithm adjusted --adjusted-statefile etc/adjusted/adjbou_state_HE_.json --interval second --starttime 2020-05-20T20:00:00 --endtime 2020-05-20T20:10:00
```
...it fails because it thinks there are no defaults. It's not the algorithm failing, but rather the Edge factory's `get_timeseries()`, because no channels are defined.

I'd swear that once-upon-a-time, default input channels were set in the Controller, but if you look at EdgeFactory's initialization, you'll see that channels are set to `None` by default, which is then passed to the initializer for TimeSeriesFactory. And because configure gets called after the factory initializations, **it doesn't seem like its actually possible to use this method for setting default inputs**.

Honestly, we might want to ping @jmfee-usgs  on this one. I'm sure I'm missing something obvious, and I'm sure he will point us in the right direction quickly.

geomagio/algorithm/AdjustedAlgorithm.py
```diff
@@ -21,24 +21,28 @@ def __init__(
         statefile=None,
         data_type=None,
         location=None,
+        inchannels=["H", "E", "Z", "F"],
```
jmfee-usgs, line 24: The defaults need to be handled differently, in case a user passes `None`.  So set the parameter default to `None`, and before calling `Algorithm.__init__`, set defaults:
```
inchannels = inchannels or ["H", "E", "Z", "F"]
...
```

@erigler-usgs : this should fix the comment you made because, without specifying channels on the input, `None` is being passed for in and out channels.

geomagio/Controller.py
```diff
@@ -152,8 +152,8 @@ def run(self, options, input_timeseries=None):
             already read the input to confirm data can be produced.
         """
         algorithm = self._algorithm
-        input_channels = options.inchannels or algorithm.get_input_channels()
-        output_channels = options.outchannels or algorithm.get_output_channels()
+        input_channels = options.inchannels or algorithm.inchannels
```
erigler-usgs, line 155: I'm worried that modifying the Controller directly this way will break other things, even if it accomplishes what we want for the Adjusted algorithm.

I believe the way this was intended to work is that methods inside the AdjustedAlgorithm class, called `get_input_channels()` and `get_output_channels()`, would return the desired default input and output channels if nothing else was passed. That is:

```python

def get_input_channels(self):
    return self.inchannels

def get_output_channels(self):
    return self.outchannels

```

But, **I don't think you even need to create these methods** because the AdjustedAlgorithm class initialization calls the Algorithm class initialization with `inchannels` and `outchannels` specified as keyword arguments, which, in turn, does exactly what I just described above.

However, even following all my own logic, things still didn't work. The reason why is due to the fact that in the `AdjustedAlgorthm.configure()` method, there is a call back to `Algorithm.configure()`, which resets `inchannels` and `outchannels` to their original command-line values (i.e., `None`). That was  a tricky bug to track down.

So, what I think needs to happen is this:

1. revert the Controller to it's original form;
2. remove the logic in `AdjustedAlgorithm.configure()` that processes the inchannels and outchannels;
3. keep what you have in `AdjustedAlgorithm.__init_()`;
4. remove the line in `AdjustedAlgorithm.configure()` that calls `Algorithm.configure()`.

When I did this locally, everything worked swimmingly.

jmfee-usgs: I'm concerned about removing the parent method calls, because we may add other functionality in the future.  This will require more thought.

geomagio/algorithm/AdjustedAlgorithm.py
```diff
@@ -233,10 +234,15 @@ def configure(self, arguments):
...
         self.load_state()
 
-        if arguments.inchannels is None:
-            self.inchannels = ["H", "E", "Z", "F"]
-            self.outchannels = ["X", "Y", "Z", "F"]
-
-        else:
-            self.inchannels = arguments.inchannels
-            self.outchannels = arguments.outchannels or arguments.inchannels
+        self._inchannels = self.inchannels = arguments.inchannels or [
```
jmfee-usgs, line 237: I think this should be fixed in the Algorithm parent class, where it parses arguments

https://github.com/usgs/geomag-algorithms/blob/ca0f5740ca8831d8976c76a2e19912f41950e6c8/geomagio/algorithm/Algorithm.py#L142

Something like this?

```
self._inchannels = arguments.inchannels or self._inchannels
self._outchannels = arguments.outchannels or self._outchannels or self._inchannels
```

geomagio/algorithm/AdjustedAlgorithm.py
```diff
@@ -37,6 +37,7 @@ def __init__(
         self.location = location
         self.inchannels = inchannels
```
jmfee-usgs, line 38: Not sure if these attributes are needed in the adjusted algorithm, since they're already in the parent class?

geomagio/algorithm/AdjustedAlgorithm.py
```diff
@@ -141,22 +131,28 @@ def process(self, stream):
...
-
-        x = self.create_trace("X", h.stats, adj[0])
-        y = self.create_trace("Y", e.stats, adj[1])
-        z = self.create_trace("Z", z.stats, adj[2])
-        f = self.create_trace("F", f.stats, fnew)
-
-        out = Stream([x, y, z, f])
+        inchannels = self.get_input_channels()
+        outchannels = self.get_output_channels()
+        # Gather input traces in order of user input(inchannels)
```
jmfee-usgs, line 136: I would streamline the input, and separate handling for F:
```
# prepare inputs
raws = np.vstack(
    [
        stream.select(channel=channel)[0].data
        for channel in inchannels if channel != "F"
    ]
    # add row of ones to inputs
    + [np.ones_like(stream[0].data)]
)
# adjust
adjusted = np.matmul(self.matrix, raws)
# prepare outputs
out = Stream([
    self.create_trace(
        outchannels[i],
        stream.select(channel=inchannels[i])[0].stats,
        adjusted[i],
    )
    for i in range(len(adjusted)-1)
])
# handle f
if "F" in inchannels and "F" in outchannels:
    f = stream.select(channel="F")[0]
    out += self.create_trace("F", f.stats, f.data + self.pier_correction)
# done
return out
```

geomagio/algorithm/AdjustedAlgorithm.py
```diff
@@ -199,6 +199,17 @@ def can_produce_data(self, starttime, endtime, stream):
         ):
             return True
 
+        # If being used for another conversion, check if all channels can produce data
```
jmfee-usgs, line 202: This can be combined with the HEZ check, just exclude the F channel.
Also, you can use `super().can_produce_data` now that we're in python 3:

```
if np.all(
    [
        super.can_produce_data(starttime, endtime, stream.select(channel=c))
        for c in channels
        if c != "F"
    ]
):
    return True
```

geomagio/algorithm/AdjustedAlgorithm.py
```diff
@@ -21,24 +21,29 @@ def __init__(
...
+
         if matrix is None:
             self.load_state()
 
     def load_state(self):
         """Load algorithm state from a file.
         File name is self.statefile.
         """
-        self.matrix = np.eye(4)
+        self.matrix = np.eye(len(self.get_input_channels()))
```
jmfee-usgs, line 46: Since this varies depending on whether it's called with F, this will be a little more complicated:

Maybe this should be also be used to load the matrix from state, since a size is expected based on the channels?

```
matrix_size = len([c for c in self.get_input_channels() if c != "F"]) + 1
self.matrix = np.eye(matrix_size)
```

geomagio/algorithm/AdjustedAlgorithm.py
```diff
@@ -21,79 +21,75 @@ def __init__(
...
     def load_state(self):
         """Load algorithm state from a file.
         File name is self.statefile.
         """
-        self.matrix = np.eye(4)
+        # Adjusted matrix defaults to identity matrix
+        matrix_size = len([c for c in self.get_input_channels() if c != "F"]) + 1
+        self.matrix = np.eye(matrix_size)
         self.pier_correction = 0
+
```
jmfee-usgs, line 50: Add a comment if you want to separate sections, generally don't leave blank lines.

```
# set defaults
...
# read statefile
if self.statefile is None:
    return
data = None
try:
    ...
if data is None or data == "":
    return
# set state
```

geomagio/algorithm/AdjustedAlgorithm.py
```diff
@@ -21,79 +21,75 @@ def __init__(
...
-        self.matrix[3, 0] = np.float64(data["M41"])
-        self.matrix[3, 1] = np.float64(data["M42"])
-        self.matrix[3, 2] = np.float64(data["M43"])
-        self.matrix[3, 3] = np.float64(data["M44"])
-        self.pier_correction = np.float64(data["PC"])
+
+        PC = data.pop("PC")
+        self.pier_correction = np.float64(PC)
+        # excludes PC
+        keys = list(data.keys())
```
jmfee-usgs, line 69: can you use `matrix_size` here (matrix is already np.eye...)?

```
for row in range(matrix_size):
    for col in range(matrix_size):
        self.matrix[row, col] = np.float64(data[f"M{row+1}{col+1}"])
```

geomagio/algorithm/AdjustedAlgorithm.py
```diff
@@ -172,34 +160,28 @@ def can_produce_data(self, starttime, endtime, stream):
...
+                for c in channels
+                if c != "F"
+            ]
         ):
+            print("true")
             return True
 
-        # return false if cannot produce adjustded F or XYZ
+        # return false if F or remaining channels cannot produce data
+        print("false")
```
jmfee-usgs, line 184: Remove the `print`s

test/algorithm_test/AdjustedAlgorithm_test.py
```diff
@@ -54,3 +54,43 @@ def test_process():
...
     assert_almost_equal(f.data, f_adj.data, 2)
+
+
+def test_process_reverse_polarity():
+    """algorithm_test.AdjustedAlgorithm_test.test_process()
+
+    Check adjusted data processing versus files generated from
+    original script. Tests reverse polarity martix.
+    """
+    matrix = None
```
jmfee-usgs, line 65: matrix and pier_correction aren't needed here, use keyword parameters instead of positional parameters:

```
a = AdjustedAlgorithm(statefile="etc/...", inchannels=["H", "E"], outchannels=["H", "E"])
```

test/algorithm_test/AdjustedAlgorithm_test.py
```diff
@@ -54,3 +54,43 @@ def test_process():
...
+    # load adjusted data transform matrix and pier correction
+    a = adj(
+        matrix,
+        pier_correction,
+        "etc/adjusted/adjbou_state_HE_.json",
+        inchannels=["H", "E"],
+        outchannels=["H", "E"],
+    )
+
+    # load boulder May 20 files from /etc/ directory
```
jmfee-usgs, line 76: name and group these a little differently, always read files in a with block so they are closed

```
with open("etc/adjusted/BOU202005vmin.min") as f:
    raw = IAGA2002Factory().parse_string(f.read())
with open("etc/adjusted/BOU202005adj.min") as f:
    expected = IAGA2002Factory().parse_string(f.read())
```

test/algorithm_test/AdjustedAlgorithm_test.py
```diff
@@ -54,3 +54,43 @@ def test_process():
...
+    he_inv_iaga2002_file = open("etc/adjusted/BOU202005adj.min")
+    he_inv_iaga2002_string = he_inv_iaga2002_file.read()
+    factory = i2.IAGA2002Factory()
+    he = factory.parse_string(he_iaga2002_string)
+    he_inv = factory.parse_string(he_inv_iaga2002_string)
+
+    # process hezf (raw) channels with loaded transform
+    adj_bou = a.process(he)
+
+    # unpack channels from loaded adjusted data file
```
jmfee-usgs, line 88: here you can select and assert in one pass:
```
assert_almost_equal(
    actual=adj_bou.select(channel="H")[0].data,
    desired=expected.select(channel="H")[0].data,
    decimal=2)
...
```

test/algorithm_test/AdjustedAlgorithm_test.py
```diff
@@ -25,7 +25,7 @@ def test_process():
     matrix = None
     pier_correction = None
     # load adjusted data transform matrix and pier correction
-    a = adj(matrix, pier_correction, "etc/adjusted/adjbou_state_.json")
+    a = adj(matrix, pier_correction, "etc/adjusted/adjbou_state_.json",)
```
jmfee-usgs, line 28: Watch for extra commas.  Not hurting anything, but also not needed.

pcain-usgs:
Black will auto-format the AdjustedAlgorithm initialization(adds comma) in test_reverse_polarity due to the added inchannels/outchannels parameters.

PR #305, opened 2020-05-26
Simpler controller refactor

jmfee-usgs:
Replaces #299


erigler-usgs:
@jmfee-usgs ,

First, a general comment:

I'm a little concerned about implementing the new processing class as part of geomag-algorithms, and not a part of a production-specific project. While geomag-algorithms is not as widely used as we may have once hoped, it remains nicely isolated from most USGS-specific operational considerations, and therefore (IMO) more useful as a community tool.

Maybe more to the point, I worry that by incorporating USGS-specific processing pipelines into geomag-algorithms, we (USGS-Geomag) will lose operational flexibility, since presumably, we'll wish to NOT break existing functionality of a community tool. It seems to me that, given what you have so far, it would be relatively easy to pull the processing class out of geomag-algorithms, and place it into a separate production-specific project. Your thoughts?

Now, more specific comments: 

It seems like all processing for an individual observatory would be done under a single Python process. Would this still be called from a shell script processing a list of observatories? Is that how we will distribute the load across multiple (but fewer) parallel processes (OS-controlled load-balancing, as opposed to using Python's thread/threading or multiprocessing modules)?

Also, it does not appear that the new processing package currently handles multi-observatory algorithms, like Dst. Or that it can handle IO factories other than Edge and Miniseed. Do we still need to do run these separately using the legacy CLI?

Finally, typer looks like a very useful library, and could potentially replace parse_args for all CL processing. I'm glad you chose not to take that leap here, since it would be an even bigger change, but do you anticipate doing something like this one day?

jmfee-usgs:
@erigler-usgs :

I think it's beneficial to publish the configurations used for processing, as that shows other users how we configure the algorithms.  We're only committing to keep the api (until the next major version) and we should be free to change the inner workings as our processing changes; this just reflects what runs on our systems.

I removed the main method as requested for now, although in the long term this should also be made available (transparency, etc).  One less moving piece if it's all in one library.  This doesn't include the configuration or cron scheduling aspects, which are separate for other reasons.


I added a basic form of the average algorithm for multi-observatory, but that will still need to be run from a separate script.  Yes, each observatory will run separately, and then the multi observatory process can run to average those outputs.  We should probably still only run the obsrio filtering every minute and leave the other processing at every 5 minutes.

Since the main script is not longer here, please take a look at the individual algorithms and let me know if you see any issues.  Once this is merged I can start the main scripts in the other projects.

erigler-usgs:
To be clear, I was only trying to initiate a discussion, I wasn't asking that the main script be removed. All things being equal, I do feel like it is better to segregate "community tools" from production tools/configurations, but honestly, that ship may have already sailed, what with the new data service, planned metadata service, and other things that we seem to be integrating with geomag-algorithms. But if we are going to keep growing geomag-algorithms in this way, I would like to come up with a plan for keeping these disparate aspects of the package clearly delineated, both in the structure of the package, and in documentation.

That all said, I see no obvious issues with the algorithms in processing, including your recent addition of the multi-observatory average algorithm for Dst. I didn't actually try to run any of them, since I don't have a local Edge to test against, but if you are confident that they will work as intended, I see no reason not to merge, and start preparing for deployment to Dev. I expect there will be issues that come up then that will need to be addressed before going to production.

erigler-usgs:
I'm happy to review and test the new IMAGCDF factory if it's ready. Or would you (@jmfee-usgs) rather I wait?

PR #306, opened 2020-06-01
Fix TimeseriesUtility.pad_and_trim_trace

erigler-usgs:
I ran into a floating point precision issue when padding a trace out
to the requested starttime when the data block returned by the Edge
miniseed factory (but this is not an Edge issue, per se) was less
than what was requested.

My solution was to use `round()` instead of `int()` when converting
a time differential to a discrete number of samples. I'm confident
this is a robust fix for padding. I'm not sure if it also the proper
approach to the trimming, which seems to be a bit more sophisticated
in that is uses the `ceil()` method. In any case, I request some
feedback before merging this deceptively simple PR.

### DIFF
--- a/geomagio/TimeseriesUtility.py
+++ b/geomagio/TimeseriesUtility.py
@@ -514,7 +514,7 @@ def pad_and_trim_trace(trace, starttime, endtime):
             trace.stats.starttime = trace_starttime
     elif trace_starttime > starttime:
         # pad to starttime
-        cnt = int((trace_starttime - starttime) / trace_delta)
+        cnt = round((trace_starttime - starttime) / trace_delta)
         if cnt > 0:
             trace.data = numpy.concatenate(
                 [numpy.full(cnt, numpy.nan, dtype=numpy.float64), trace.data]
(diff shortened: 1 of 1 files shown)


erigler-usgs:
While not completely reproducing the situation this PR is designed to fix, if you simply type `(7.1 - 6.4)/0.1` into a Python interpreter, you will quickly see the problem (ans: 6.999999999999993). If you then apply the `int()` function to this, you get `6`, something that is certainly NOT what is wanted.

jmfee-usgs:
@erigler-usgs : this definitely requires some caution.

I think it uses `int` to find the nearest sample that doesn't fall outside `[starttime,endtime]`.  This also breaks the unit tests that are checking this logic.  It's worth adding a unit test to reproduce the existing problem.

A safer approach might be using an epsilon value to deal with the precision issue, something like 
`int( ((trace_starttime - starttime) / trace_delta) + 1e-6 )`

erigler-usgs:
@jmfee-usgs , I pushed a second commit before reading your response. I took a slightly different approach to the "epsilon" suggestion, but it seems to work well, if not quite as efficiently.

I'll update for the "trim" logic as well, and work on that unit test. Stay tuned.

erigler-usgs:
@jmfee-usgs ...just a quick ping to ask you to approve/merge. Thx.

jmfee-usgs:
@erigler-usgs hoping to find time for this tomorrow

PR #307, opened 2020-06-02
Residual diagnostics

pcain-usgs:
Adds diagnostics in residual calculations currently seen in the observation webservice.

### DIFF
--- a/geomagio/residual/Calculation.py
+++ b/geomagio/residual/Calculation.py
@@ -13,6 +13,7 @@ from .MeasurementType import (
     MARK_TYPES,
 )
 from .Measurement import AverageMeasurement, Measurement, average_measurement
+from .Diagnostics import Diagnostics
 from .Reading import Reading
 
 
@@ -42,7 +43,7 @@ def calculate(reading: Reading, adjust_reference: bool = True) -> Reading:
         mean=mean,
         reference=adjust_reference and reference or None,
     )
-    absoluteD = calculate_D_absolute(
+    absoluteD, average_mark, meridian, azimuth, mark_azimuth = calculate_D_absolute(
         azimuth=reading.azimuth,
         h_baseline=absoluteH.baseline,
         measurements=reading.measurements,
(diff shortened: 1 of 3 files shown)


geomagio/residual/Calculation.py
```diff
@@ -121,13 +150,26 @@ def calculate_D_absolute(
...
-        starttime=mean.time,
-        endtime=mean.endtime,
+
+    # for diagnostics
+    if average_mark > 180:
+        average_mark -= 90
+    if meridian > 180:
+        meridian -= 90
+
+    return (
```
jmfee-usgs, line 160: Also update the method return signature...

In this case it seems like all the diagnostics are generated in this method, so it would be better to just return a Diagnostics object.

```
def calculate_D_absolute(...) -> Tuple[Absolute, Diagnostics]:
    ...
```

geomagio/residual/Calculation.py
```diff
@@ -57,12 +58,23 @@ def calculate(reading: Reading, adjust_reference: bool = True) -> Reading:
             inclination=inclination,
             measurements=scale_measurements,
         )
+    diagnostics = Diagnostics(
+        meridian=meridian,
+        mean_mark=average_mark,
+        magnetic_azimuth=magnetic_azimuth,
+        mark_azimuth=azimuth,
+        declination=absoluteD.absolute,
```
jmfee-usgs, line 66: Calculated values that are already returned should not be duplicated in diagnostics.

geomagio/residual/Calculation.py
```diff
@@ -113,6 +125,23 @@ def calculate_D_absolute(
             for m in declination_measurements
         ]
     )
+    # calculate south-facing meridian for diagnostics
+    south_meridian = np.average(
```
jmfee-usgs, line 129: This isn't just an average of measurements?  Is there a reference for this calculation?

geomagio/residual/SpreadsheetAbsolutesFactory.py
```diff
@@ -9,6 +9,8 @@
 from .Measurement import Measurement
 from .MeasurementType import MeasurementType as mt
 from .Reading import Reading
+from .Diagnostics import Diagnostics
```
jmfee-usgs, line 12: Watch import alphabetical order.

geomagio/residual/SpreadsheetAbsolutesFactory.py
```diff
@@ -403,6 +406,25 @@ def _parse_metadata(
...
+        """
+        Gather diagnostics from list of measurements
+        """
+        mean_mark = average_measurement(measurements, MARK_TYPES).angle
+
+        meridian = average_measurement(measurements, DECLINATION_TYPES).angle
+
+        magnetic_azimuth = mean_mark - meridian
+
+        if meridian > 180:
```
jmfee-usgs, line 419: are these adjustments needed here?
and, if so, should mean_mark be adjusted first?

Also, remove the blank lines

geomagio/residual/Diagnostics.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Diagnostics(BaseModel):
+    """Computed diagnostics during calculation.
+
+    Attributes
+    ----------
+    meridian: claculated from declination measurements
+    mean_mark: average mark angles from measurements
+    magnetic_azimuh: after adjustment
```
jmfee-usgs, line 13: watch spelling, and remove the no longer used attribute comments.

PR #311, opened 2020-06-10
Fixes Issue 310

pcain-usgs:
Edit to FilterAlgorithm that raises a ValueError with bad starttimes in traces.


pcain-usgs:
> What do you think about this?
> 
> ```
> 
> # first output timestamp is in the center of the filter window
> filter_time_shift = input_sample_period * (numtaps // 2)
> # data to filter
> data = trace.data
> starttime = trace.stats.starttime + filter_time_shift
> # align with the output period
> misalignment = starttime.timestamp % output_sample_period
> if misalignment != 0:
>     # skip incomplete input
>     starttime = (starttime - misalignment) + output_sample_period
>     input_starttime = starttime - filter_time_shift
>     offset = int(1e-6 + (input_starttime - trace.stats.starttime) / input_sample_period)
>     print(f"Input misaligned {misalignment}s, skipping {offset} input samples", file=sys.stderr)
>     data = data[offset:]
>     # check there is still enough data for filter
>     if len(data) < numtaps:
>         continue
> ```

This is similar to an edit I have been working on, but I have placed this process in a separate method with a while loop(loop until misalignment is 0). Let me know if you think the better path is to place this directly into the process method

jmfee-usgs:
I'm not sure about the loop?  Shouldn't misalignment be zero after the shift?

pcain-usgs:
> I'm not sure about the loop? Shouldn't misalignment be zero after the shift?
After looking more closely, I see that this will move the misalignment to zero after one shift. Disregard the comment about looping. Should this be in its own method?

jmfee-usgs:
Implementers choice.  Maybe `get_aligned_input(trace, input_period, output_period, numtaps) -> Tuple[numpy.array, UTCDateTime]` to return the aligned input data and starttime?

jmfee-usgs:
Please add a unit test to verify this works as expected

test/algorithm_test/FilterAlgorithm_test.py
```diff
@@ -171,3 +171,31 @@ def test_custom():
...
+    """
+    f = FilterAlgorithm(input_sample_period=1.0, output_sample_period=60.0,)
+    with open("etc/filter/BOU20200101vsec.sec", "r") as file:
+        iaga = i2.StreamIAGA2002Factory(stream=file)
+        bou = iaga.get_timeseries(starttime=None, endtime=None, observatory="BOU")
+    filtered = f.process(bou)
+
+    with open("etc/filter/BOU20200101_filtered_vsec.sec", "r") as file:
+        iaga = i2.StreamIAGA2002Factory(stream=file)
+        BOU = iaga.get_timeseries(starttime=None, endtime=None, observatory="BOU")
```
jmfee-usgs, line 188: Use a different variable name, maybe `bou_expected`?

Also, this test is less about verifying the filtered values match (that's a different test), and more about testing alignment of the filtered values.

test/algorithm_test/FilterAlgorithm_test.py
```diff
@@ -171,3 +171,31 @@ def test_custom():
...
+
+def test_starttime_shift():
+    """algorithm_test.FilterAlgorithm_test.test_starttime_shift()
+    Tests algorithm for second to minute with misalligned starttime(1 second).
+    """
+    f = FilterAlgorithm(input_sample_period=1.0, output_sample_period=60.0,)
+    with open("etc/filter/BOU20200101vsec.sec", "r") as file:
+        iaga = i2.StreamIAGA2002Factory(stream=file)
+        bou = iaga.get_timeseries(starttime=None, endtime=None, observatory="BOU")
+    filtered = f.process(bou)
```
jmfee-usgs, line 184: Can you make an assertion about expected time of the first filtered sample?
What about trimming the input to different times, and getting the same filtered outputs (or similar)?

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -67,12 +68,22 @@ def load_state(self):
         if data is None or data == "":
             return
 
+        window = data["window"]
+
+        if len(window) % 2 == 0:
```
jmfee-usgs, line 73: Move this to a separate method (`_prepare_step(step) -> Dict`?) so all steps are checked regardless of how a user configures them.

Print a message to stderr if the window was changed.
Return a new dictionary if changes are made, so the original (input) step is unchanged.
Use numpy.insert:
```
if len(step["window"]) % 2 == 1:
    return step
new_window = numpy.array(window)
i = len(window) // 2
numpy.insert(new_window, i + 1, numpy.average(window[i:i+2]))
new_step = dict(step)
new_step["window"] = new_window
return new_step
```

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -188,9 +199,32 @@ def process_step(self, step, stream):
 
         out = Stream()
         for trace in stream:
-            filtered = self.firfilter(trace.data, window, decimation)
             stats = Stats(trace.stats)
```
jmfee-usgs, line 202: Wait to create a new `Stats` until after the data are filtered below.

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -188,9 +199,32 @@ def process_step(self, step, stream):
...
-            stats.starttime = stats.starttime + input_sample_period * (numtaps // 2)
+            # first output timestamp is in the center of the filter window
+            filter_time_shift = input_sample_period * (numtaps // 2)
```
jmfee-usgs, line 204: This can be grouped with the other step variables outside the `for` loop

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -188,9 +199,32 @@ def process_step(self, step, stream):
...
+            misalignment = starttime.timestamp % output_sample_period
+            if misalignment != 0:
+                # skip incomplete input
+                starttime = (starttime - misalignment) + output_sample_period
+                input_starttime = starttime - filter_time_shift
+                offset = int(
+                    1e-6
+                    + (input_starttime - trace.stats.starttime) / input_sample_period
+                )
+                print(
```
jmfee-usgs, line 218: Maybe the time is less important than number of samples.

```
print(f"Skipping {offset} input samples to align output", file=sys.stderr)
```

test/algorithm_test/FilterAlgorithm_test.py
```diff
@@ -171,3 +171,32 @@ def test_custom():
...
+
+def test_starttime_shift():
+    """algorithm_test.FilterAlgorithm_test.test_starttime_shift()
+    Tests algorithm for second to minute with misalligned starttime(16 seconds).
+    """
+    f = FilterAlgorithm(input_sample_period=1.0, output_sample_period=60.0,)
+    # generation of BOU20200101vsec.sec
+    # starttime = UTCDateTime('2020-01-01T00:00:00Z')
+    # endtime = UTCDateTime('2020-01-01T00:15:00Z')
+    # starttime,endtime = f.get_input_interval(starttime,endtime)
```
jmfee-usgs, line 184: Rather than use the filter algorithm to prepare input data, just fetch a fixed interval of data ('2020-01-01T00:00:00Z' - '2020-01-01T00:15:00Z').  Since you know the interval you can:

- filter based on available data, should produce 00:01:00-00:14:00
- trim the input to precisely the input data (00:00:15-00:14:45), should produce the same (00:01:00-00:14:00)
- trim slightly more (00:00:16-00:14:44), should produce less (00:02:00-00:13:00)

etc/filter/coeffs.json
```diff
@@ -1 +1 @@
...
...6820028521606, 0.0007960763923794446, -0.005632424567553352, -0.010350920336899747, -0.013356481033429665, -0.014751069175579008, -0.014724611301843292, -0.013532939805804448, -0.01147277377194849, -0.008856003607624379, -0.005985429793316389, -0.003133805469446045, -0.0005275870204826471, 0.0016637408771624203, 0.0033324707796725693, 0.004431861432275873, 0.004970436385096754, 0.005002478119825619, 0.0046160248379807445, 0.003919751576353979, 0.003030052274417014, 0.0020594556416456715, 0.0011072361616178837, 0.0002527592710099409, -0.00044823496583773606, -0.0009645169835343994, -0.00128778834564552, -0.001429191565092709, -0.0014145673052782036, -0.0012791392276245408, -0.0010622553823292297, -0.0008027060600497556, -0.0005349914806186782, -0.00028675003259182703, -7.739918781232675e-05, 8.209652210988856e-05, 0.00018852107781718695, 0.0002450226563246427, 0.000259332095242354, 0.00024185700619301848, 0.0002038861691000964, 0.0001560816599615354, 0.00010737006066026785, 6.427827433229639e-05, 3.0701315868917085e-05, 8.043963361551653e-06, -4.352200290023471e-06, -8.59990999152696e-06, -7.5805261904864375e-06, -4.277196717501353e-06, -1.2056376791654907e-06, 7.2383512291811e-20]}
```
jmfee-usgs, line 1: With all the values on one line, it's hard to tell what changed.  Use an autoformatter to make this easier to read.

geomagio/algorithm/FilterAlgorithm.py
```diff
@@ -2,6 +2,7 @@
 from .Algorithm import Algorithm
 import numpy as np
 import scipy.signal as sps
+import sys
```
jmfee-usgs, line 5: Organize the imports (builtins, packages, locals)

geomagio/algorithm/FilterAlgorithm.py
jmfee-usgs: Add a test for this.

You could revert changes to the custom filter, and have the test explicitly edit the filter.

jmfee-usgs:
rebased and merged externally with fixes

